PR #2, opened 2023-06-28
add initial .gitlab-ci.yaml

SimeonEhrig:
Add also a GitHub Action to accelerate the start of the GitLab CI. We use the official [GitHub mirror to Gitlab CI](https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html) feature of gitlab.com. 

This feature works quit well, but it has a big disadvantage. gitlab.com is looking from time to time, if something changed in your GitHub repository and if yes, it triggers the mirror process. The check interval depends on the server utilization of gitlab.com. Therefore a CI start can vary from a few seconds until some minutes.

We solved the problem, with a GitHub Action jobs. The job starts directly after pushing a commit or open a pull request and simply push the changes to gitlab.com. This triggers the CI and gitlab creates the CI link in the pull request. With the GitHub action the CI startup time is constant about some seconds.

A positive side effect is also, that we have a GitHub Action job, which represent the GitLab CI. With this job, we can also configure a execution order between GitHub Action jobs and GitLab CI.


SimeonEhrig:
@szabo137 Do you know, why the GitHub Action is not started?

.github/workflows/gitlab-mirrow.yaml
```diff
@@ -0,0 +1,21 @@
+name: Mirror and run GitLab CI
+
+on: [push]
```
szabo137, line 3: I think here you need to include `pull_request` in the list for triggering events.

SimeonEhrig: Yes, it works. I understand also my mistake. I tested the extension before only with PR's, where both branches was located in the same repository. Therefore the push to the repo triggered the CI and not the PR.

Now, there is another problem. Let's see, if the GitHub Action can handle forks.

SimeonEhrig:
GitHub Action does not work to test forks in GitLab CI

PR #14, opened 2023-09-26
Computation and process setup interface

szabo137:
With this PR, the concept of *computation setups* is introduced, and general as well as functionality related to scattering processes is implemented. 

## Description of the problem

One of the main tasks of this package is the computation of quantities like differential and total cross-section for given scattering processes and a given set of parameters. Usually, these quantities depend on a fixed set of initial parameters, and their value needs to be calculated for large amounts of input data. For example, the differential cross-section of a process depends on the generic scattering process, the compute model, and all sorts of initial parameters, which might be physical (e.g. energy scales) or technical (e.g. integrator settings). Once initialized, the differential cross-section is still a function of the external momenta of a scattering process. Therefore, for a given set of input data, i.e. momenta, one needs to be able to compute the respective value of the differential cross-section using the given setting. 

## Suggested solution

The initial parameters for a given quantity will be collected in a *setup*, which, once initialized, describes the *initialized quantity*, i.e. the quantity with fixed initial parameters. The actual computation can be performed by calling a member function `compute` on the setup object and the respective input arguments. The implementation described below defines an interface, that unifies this approach and extends the computation workflow by adding input verification and post-processing steps. 

## Implementation details
The root type for all setups is `AbstractSetup`, for which the following interface functions are defined:

```Julia 
_input_validation(stp::AbstractSetup, input::Any)
_compute(stp::AbstractSetup, input::Any)
_post_computation(stp::AbstractSetup, input::Any, result::Any)
```

None of those functions is exported, but they need to be added for a concrete implementation of `AbstractSetup`. For all functions, except `_compute`, a generic fallback is implemented, which uses a default implementation for the respective function. Based on the interface functions, the actual compute function is implemented as 

```Julia

compute(stp::AbstractSetup, input::Any)

```
where internally, the following steps are performed:

1. input validation by calling `_input_validation` 
2. actual computation by calling `_compute`
3. post-processing by calling `_post_computation`

Additionally, based on `AbstractSetup`, a specialized version of setups related to the combination of scattering processes and compute models is implemented: `AbstractProcessSetup<:AbstractSetup`, where the following functions are added to the setup interface:

```Julia
scattering_process(stp::AbstractProcessSetup)
compute_model(stp::AbstractProcessSetup)
```

Based on them, the following functions are delegated to the respective type parameter:

```Julia
number_incoming_particles(stp::AbstractProcessSetup)
number_outgoing_particles(stp::AbstractProcessSetup)
```

## Final remarks

The possibilities for functionality based on the setup definition above are not exhausted by the set of interface functions defined with this PR. For example, the pre-and post-processing could be enhanced by more fine granular structures. Furthermore, the initialization step of a setup could have its input validation. This is currently delegated to the users, who implement a setup by themself. Finally, the set of functions defined on `AbstractProcessSetup` and delegated to the process and model, could be extended if necessary. However, all the points above should be considered only if they are necessary, which implies opening dedicated issues if any.

### DIFF
--- a/src/interfaces/process_interface.jl
+++ b/src/interfaces/process_interface.jl
@@ -121,10 +121,10 @@ function differential_cross_section(
     final_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
     ) where {T<:QEDbase.AbstractFourMomentum}
     size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
     )
     size(final_phasespace, 1) == number_outgoing_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_pariticles(proc_def)}>."),
+        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_particles(proc_def)}>."),
     )
     return _differential_cross_section(proc_def, model_def, init_phasespace, final_phasespace)
 end
@@ -257,7 +257,7 @@ function total_cross_section(
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
 ) where {T<:QEDbase.AbstractFourMomentum}
     size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
     )
     return _total_cross_section(proc_def, model_def, init_phasespace)
 end
--- a/src/interfaces/process_interface.jl
+++ b/src/interfaces/process_interface.jl
@@ -115,14 +115,23 @@ function differential_cross_section(
     model_def::AbstractModelDefinition,
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
     final_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
-    ) where {T<:QEDbase.AbstractFourMomentum}
+) where {T<:QEDbase.AbstractFourMomentum}
     size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
+        DimensionMismatch(
+            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>.",
+        ),
     )
     size(final_phasespace, 1) == number_outgoing_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_particles(proc_def)}>."),
+        DimensionMismatch(
+            "The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_particles(proc_def)}>.",
+        ),
     )
-    return _differential_cross_section(proc_def, model_def, init_phasespace, final_phasespace)
+    return _differential_cross_section(
+        proc_def,
+        model_def,
+        init_phasespace,
+        final_phasespace,
+    )
 end
 
 # returns diffCS for single `initPS` and several `finalPS` points without input-check
@@ -253,7 +262,9 @@ function total_cross_section(
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
 ) where {T<:QEDbase.AbstractFourMomentum}
     size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
-        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
+        DimensionMismatch(
+            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>.",
+        ),
     )
     return _total_cross_section(proc_def, model_def, init_phasespace)
 end
--- a/src/interfaces/process_interface.jl
+++ b/src/interfaces/process_interface.jl
@@ -47,7 +47,7 @@ function outgoing_particles end
 
 Return the number of incoming particles of a given process. 
 """
-@inline function number_incoming_pariticles(proc_def::AbstractProcessDefinition)
+@inline function number_incoming_particles(proc_def::AbstractProcessDefinition)
     return length(incoming_particles(proc_def))
 end
 
@@ -57,7 +57,7 @@ end
 
 Return the number of outgoing particles of a given process. 
 """
-@inline function number_outgoing_pariticles(proc_def::AbstractProcessDefinition)
+@inline function number_outgoing_particles(proc_def::AbstractProcessDefinition)
     return length(outgoing_particles(proc_def))
 end
 
@@ -119,23 +119,14 @@ function differential_cross_section(
     model_def::AbstractModelDefinition,
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
     final_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
-) where {T<:QEDbase.AbstractFourMomentum}
-    size(init_phasespace, 1) == number_incoming_pariticles(proc_def) || throw(
-        DimensionMismatch(
-            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>.",
-        ),
+    ) where {T<:QEDbase.AbstractFourMomentum}
+    size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
     )
-    size(final_phasespace, 1) == number_outgoing_pariticles(proc_def) || throw(
-        DimensionMismatch(
-            "The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_pariticles(proc_def)}>.",
-        ),
+    size(final_phasespace, 1) == number_outgoing_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_pariticles(proc_def)}>."),
     )
-    return _differential_cross_section(
-        proc_def,
-        model_def,
-        init_phasespace,
-        final_phasespace,
-    )
+    return _differential_cross_section(proc_def, model_def, init_phasespace, final_phasespace)
 end
 
 # returns diffCS for single `initPS` and several `finalPS` points without input-check
@@ -265,10 +256,8 @@ function total_cross_section(
     model_def::AbstractModelDefinition,
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
 ) where {T<:QEDbase.AbstractFourMomentum}
-    size(init_phasespace, 1) == number_incoming_pariticles(proc_def) || throw(
-        DimensionMismatch(
-            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>.",
-        ),
+    size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
     )
     return _total_cross_section(proc_def, model_def, init_phasespace)
 end
--- a/src/interfaces/process_interface.jl
+++ b/src/interfaces/process_interface.jl
@@ -3,10 +3,6 @@
 #
 # In this file, we define the interface for working with scattering processes in
 # general.
-# 
-# This file is part of `QEDprocesses.jl` which is by itself part of the `QED.jl`
-# ecosystem.
-#
 ###############
 
 """
--- a/src/interfaces/process_interface.jl
+++ b/src/interfaces/process_interface.jl
@@ -66,8 +66,8 @@ end
         final_phasespace::AbstractVector{T},
     ) where {T<:QEDbase.AbstractFourMomentum}
 
-Interface function for the combination of scattering processes and models. Return the differential cross section of a 
-given process and model for a passed initial and final phase space. The elements of the `AbstractVector` representing the phase spaces 
+Interface function for the combination of scattering processes and physical models. Return the differential cross section of a 
+given process and physical model for a passed initial and final phase space. The elements of the `AbstractVector` representing the phase spaces 
 are the momenta of the respective particles. The implementation of this function for a concrete process and model must not 
 check if the length of the passed phase spaces match the respective number of particles. 
 
@@ -194,7 +194,7 @@ end
         init_phasespace::AbstractVector{T},
     ) where {T<:QEDbase.AbstractFourMomentum} end
 
-Interface function for the combination of scattering processes and models. Return the total cross section of a 
+Interface function for the combination of scattering processes and physical models. Return the total cross section of a 
 given process and model for a passed initial phase space. The elements of the `AbstractVector` representing the initial phase space
 are the momenta of the respective particles. The implementation of this function for a concrete process and model must not 
 check if the length of the passed initial phase spaces match number of incoming particles. 
@@ -242,7 +242,7 @@ end
         init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
     ) where {T<:QEDbase.AbstractFourMomentum}
 
-Return the total cross section for a combination of a scattering process and a compute model evaluated on a given initial phase space. 
+Return the total cross section for a combination of a scattering process and a physical model evaluated on a given initial phase space. 
 
 This function will eventually call the respective interface function [`_total_cross_section`](@ref).
 
--- a/src/interfaces/setup_interface.jl
+++ b/src/interfaces/setup_interface.jl
@@ -44,15 +44,15 @@ quantity depends on is kept constant.
     which computes the value of the associated quantity for a given `input` (see [`_compute`](@ref) for more details).
 
 
-    ## Post computation
+    ## Post processing 
 
     Every subtype of `AbstractComputationSetup` should implement the interface function
     
     ```Julia
-    _post_computation(stp::AbstractComputationSetup, input) 
+    _post_processing(stp::AbstractComputationSetup, input, result) 
     ```
     
-    which performs *computations* after the actual computation, e.g. conversions or normalizations (see [`_post_computation`](@ref) for more details).
+    which performs task after the actual computation, e.g. conversions or normalizations (see [`_post_processing`](@ref) for more details).
 
 
     
@@ -120,17 +120,17 @@ end
 
 """
     
-    function _post_computation(stp::AbstractComputationSetup, input::Any, result::Any)
+    function _post_processing(stp::AbstractComputationSetup, input::Any, result::Any)
     
 Interface function, which is called in [`compute`](@ref) after [`_compute`](@ref) has been called. This function is dedicated to 
 finalize the result of a computation. 
 
 !!! note "default implementation"
 
-    Since in the case of no post computation the result of [`_compute`](@ref) is unchanged, this function returns `result` by default.
+    Since in the case of no post processing the result of [`_compute`](@ref) is unchanged, this function returns `result` by default.
 
 """
-@inline function _post_computation(stp::AbstractComputationSetup, input, result)
+@inline function _post_processing(stp::AbstractComputationSetup, input, result)
     return result
 end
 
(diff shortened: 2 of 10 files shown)


src/interfaces/process_interface.jl
```diff
@@ -119,23 +119,14 @@ function differential_cross_section(
...
-            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>.",
-        ),
-    )
-    size(final_phasespace, 1) == number_outgoing_pariticles(proc_def) || throw(
-        DimensionMismatch(
-            "The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_pariticles(proc_def)}>.",
-        ),
+    ) where {T<:QEDbase.AbstractFourMomentum}
+    size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
```
tjungni, line 124:
```suggestion
        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
```

src/interfaces/process_interface.jl
```diff
@@ -119,23 +119,14 @@ function differential_cross_section(
...
     )
-    return _differential_cross_section(
-        proc_def,
-        model_def,
-        init_phasespace,
-        final_phasespace,
+    size(final_phasespace, 1) == number_outgoing_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_pariticles(proc_def)}>."),
```
tjungni, line 127:
```suggestion
        DimensionMismatch("The number of momenta in the final phasespace <{length(final_phasespace)}> does not match the number of outgoing particles of the process <{number_outgoing_particles(proc_def)}>."),
```

src/interfaces/process_interface.jl
```diff
@@ -265,10 +256,8 @@ function total_cross_section(
     model_def::AbstractModelDefinition,
     init_phasespace::Union{AbstractVector{T},AbstractMatrix{T}},
 ) where {T<:QEDbase.AbstractFourMomentum}
-    size(init_phasespace, 1) == number_incoming_pariticles(proc_def) || throw(
-        DimensionMismatch(
-            "The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>.",
-        ),
+    size(init_phasespace, 1) == number_incoming_particles(proc_def) || throw(
+        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_pariticles(proc_def)}>."),
```
tjungni, line 260:
```suggestion
        DimensionMismatch("The number of momenta in the initial phasespace <{length(init_phasespace)}> does not match the number of incoming particles of the process <{number_incoming_particles(proc_def)}>."),
```

tjungni (commented):
Some pariticles are still hiding

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+
+    compute(stp::AbstractComputationSetup, input::Any)
+
+Return the value of the quantity associated with `stp` for a given `input`. 
+In addition to the actual call of the associated unsafe version [`_compute`](@ref),
+input validation (using [`_input_validation`](@ref)) and post computation
+(using [`_post_computation`](@ref)) are wrapped around the calculation (see [`AbstractComputationSetup`](@ref) for details).
+"""
+function compute(stp::AbstractComputationSetup, input)
+    _input_validation(stp,input) || error("InvalidInputError: there is something wrong with the input!\n setup:$stp \n input:$input")
```
AntonReinhard, line 126:
```suggestion
    _input_validation(stp,input) || error("InvalidInputError: there is something wrong with the input!\n  setup: $stp \n  input: $input")
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+    raw_result = _compute(stp,input)
+    return _post_computation(stp, input,raw_result)
+end
+
+"""
+Abstract base type for setups related to a combination scattering processes and compute models.  
+Every subtype of `AbstractProcessSetup` is expected to implement at least the following 
+interface functions
```
AntonReinhard, line 134:
```suggestion
Abstract base type for setups related to combination scattering processes and compute models.  
Every subtype of `AbstractProcessSetup` must implement at least the following 
interface functions:
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+
+```Julia
+scattering_process(::AbstractProcessSetup) 
+compute_model(::AbstractProcessSetup) 
+```
+
+Derived from the these interface functions, the following delegations are implemented
```
AntonReinhard, line 141:
```suggestion
Derived from these interface functions, the following delegations are provided:
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+
+"""
+abstract type AbstractProcessSetup <: AbstractComputationSetup end
+
+"""
+    
+    scattering_process(stp::AbstractProcessSetup)
+
+Interface function, which returns the scattering process associated with `stp`,
+i.e. an object which is a subtype of `AbstractProcessDefinition`.
```
AntonReinhard, line 156:
```suggestion
Interface function that returns the scattering process associated with `stp`,
i.e. an object which is a subtype of `AbstractProcessDefinition`. 
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+abstract type AbstractProcessSetup <: AbstractComputationSetup end
+
+"""
+    
+    scattering_process(stp::AbstractProcessSetup)
+
+Interface function, which returns the scattering process associated with `stp`,
+i.e. an object which is a subtype of `AbstractProcessDefinition`. 
+"""
+function scattering_process end
```
AntonReinhard, line 158: As with compute, I'd at least add the parameters and a throw with an error message:
```suggestion
function scattering_process(stp::AbstractProcessSetup)
    throw("UnimplementedError: you must implement scattering_process(stp::ProcessSetup) for every subtype of AbstractProcessSetup"
end
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+
+"""
+
+    compute_model(stp::AbstractProcessSetup)
+
+Interface function, which returns the compute model associated with `stp`, i.e.
+an object which is subtype of `AbstractModelDefinition`
+"""
+function compute_model end
```
AntonReinhard, line 167:
```suggestion
function compute_model(stp::AbstractProcessSetup)
    throw("UnimplementedError: you must implement compute_model(stp::ProcessSetup) for every subtype of AbstractProcessSetup"
end
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+i.e. an object which is a subtype of `AbstractProcessDefinition`. 
+"""
+function scattering_process end
+
+"""
+
+    compute_model(stp::AbstractProcessSetup)
+
+Interface function, which returns the compute model associated with `stp`, i.e.
+an object which is subtype of `AbstractModelDefinition`
```
AntonReinhard, line 165:
```suggestion
Interface function that returns the compute model associated with `stp`, i.e.
an object which is a subtype of `AbstractModelDefinition`.
```

test/interfaces/setup_interface.jl
AntonReinhard, line 1: If we do end up using the throwing empty implementations of `_compute`, `scattering_process` and `compute_model`, those throws should probably be tested too

Misspellings across the file: 
- `thruth` => `truth`

test/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,148 @@
...
+QEDprocesses._post_computation(::TestSetupCustomPostComputationFAIL,x,y) = _groundtruth_post_computation(x,y)
+@testset "general computation setup interface" begin
+    @testset "interface fail" begin
+        rnd_input = rand(RNG)
+        @test_throws MethodError QEDprocesses._compute(TestSetupFAIL(), rnd_input) 
+        @test_throws MethodError compute(TestSetupFAIL(), rnd_input) 
+
+        @test_throws MethodError QEDprocesses._compute(TestSetupCustomValidationFAIL(), rnd_input) 
+        @test_throws MethodError compute(TestSetupCustomValidationFAIL(), rnd_input) 
+        # invalid input should be catched without throwing a MethodError
```
AntonReinhard, line 53:
```suggestion
        # invalid input should be caught without throwing a MethodError
```

test/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,148 @@
...
+    @testset "($N_INCOMING,$N_OUTGOING)" for (N_INCOMING,N_OUTGOING) in Iterators.product(
+                                                                                          (1, rand(RNG, 2:8)), (1, rand(RNG, 2:8))
+                                                                                         )   
+        INCOMING_PARTICLES = rand(RNG,PARTICLE_SET,N_INCOMING)
+        OUTGOING_PARTICLES = rand(RNG,PARTICLE_SET,N_OUTGOING)
+
+        QEDprocesses.incoming_particles(::TestProcess) = INCOMING_PARTICLES
+        QEDprocesses.outgoing_particles(::TestProcess) = OUTGOING_PARTICLES 
+
+        @testset "deligated functions" begin
```
AntonReinhard, line 141:
```suggestion
        @testset "delegated functions" begin
```

AntonReinhard (commented):
Sorry for being maybe a little pedantic :sweat_smile:

src/interfaces/setup_interface.jl
tjungni: The first sentence still doesn't sound right with the edit. Either "a combination of" or "combining"

AntonReinhard: You're right, I edited it

szabo137:
I added a description to this PR and therefore removed the draft status.

test/interfaces/setup_interface.jl
szabo137: The throwing of `MethodError` for `scattering_process` and `compute_model` is tested in the test set "interface fail" on lines 119 to 122. But you are right, of course, the `_compute` method should be tested as well. Therefore, 944885f adds the test accordingly.

Regarding the misspelling: this was a combination of *one type* and the doom of neovim's autocompletion, but d698b56 fixed this one.

szabo137 (commented):
I reacted to every concern, please feel free to comment on my replies.

szabo137:
> Sorry for being maybe a little pedantic 😅

Not at all, this review was very good and helped a lot to improve the PR!

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+
+Interface function which returns the value of the associated quantity evaluated on `input`, which can be anything the associated quantity is defined to be feasable for.
+
+!!! note "unsafe implementation"
+
+    This function should be implemented without any input validation or post computations (see [`_input_validation`](@ref) and [`_post_computation`](@ref)). The latter two are performed while calling 
+    the safe version of this function [`compute`](@ref).
+
+"""
+function _compute end
```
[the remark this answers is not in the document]
AntonReinhard: Okay this makes sense. What I had previously learned from code bases is that usually, code is better than documentation. That's why I would have preferred the type signature in the function definition rather than the docstring, but I can see the reasoning behind it now.

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,171 @@
...
+    
+"""
+abstract type AbstractComputationSetup end
+
+# convenience function to check if an object is a computation setup
+_is_computation_setup(::AbstractComputationSetup) = true
+
+"""
+
+    _is_valid_input(stp::AbstractComputationSetup, input::Any)
```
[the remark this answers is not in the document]
AntonReinhard: That sounds good, but I'd then suggest making _is_valid_input a default implementation that's just something like
```julia
function _is_valid_input(stp::AbstractComputationSetup, input)
    try
        _assert_valid_input(stp, input)
    catch e
        return false
    end
    return true
end
```

Otherwise you would have to implement two functions that do essentially the same thing every time.

tjungni (commented):
A few more typos

src/interfaces/setup_interface.jl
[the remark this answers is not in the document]
szabo137: In both cases, there are situations where you need to overwrite both functions, which was also mentioned in the docstring of `_assert_valid_input`. 

In your case, it is not possible to just overwrite `_is_valid_input` without writing an adjusted version of `_assert_valid_input`. For the current version, its similar, if you provide a custom implementation of `_assert_valid_input`, you also need to implement a custom version of `_is_valid_input`. 

Therefore, the questions comes down to: which of the two functions can be should be customized without the need to change the other. I thought, that the implementation of `_is_valid_input` is easier and more convenient than an assert function (especially if you want to use `_is_valid_input` regularly outside the assert), and should therefore take the lead. Finally, following the [Julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/#Don't-overuse-try-catch), one should avoid try-catch blocks if unnecessary, and they tend to be slow.

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,206 @@
...
+"""
+@inline function _assert_valid_input(stp::AbstractComputationSetup,input)
+    _is_valid_input(stp,input) || throw(InvalidInputError("Something wrong with the input!\n setup:$stp \n input:$input"))
+
+    return nothing
+end
+
+"""
+    
+    function _post_computation(stp::AbstractComputationSetup, input::Any, result::Any)
```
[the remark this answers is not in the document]
szabo137: Agree! 6dc2222 renamed `_post_computation` to `_post_processing`, and also replaced ever appearance of the notion *post computation* with *post processing*.

src/interfaces/setup_interface.jl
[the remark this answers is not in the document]
szabo137: As a follow-up to our offline discussion, I suggest using `physical_model`, because something a theory like quantum field theory is usually used in a general context, whereas quantum electrodynamics (one of the things indicated by `physical_model`) is more specifically bound to the Standard model of particle physics. 

Therefore, 196df3c changes `compute_model` to `physical_model`.

src/interfaces/setup_interface.jl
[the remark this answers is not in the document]
AntonReinhard: I don't follow.

If you used my suggested implementation for `_is_valid_input` I don't see any reason why you would ever need to override it at all. You'd just have to implement `_assert_valid_input` for your type and the rest will be solved by the general `_is_valid_input` implementation dispatching to the correct asserting variant, or am I misunderstanding something?

The reason why I'd suggest the assert variant to be the one to implement instead of the boolean variant is that it uses an actual error-handling mechanism to handle errors. The boolean variant can give the user no information about what happened or where, the user only gets to know "something broke". 

The style guide says "Don't *over*use try-catch", but I think one encapsulating try-catch of a function that is expected to throw sometimes is overusing. I think what the style guide is saying is that you shouldn't wrap everything in try-catch just to be safe, but that's not what we're doing.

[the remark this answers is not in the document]
szabo137: I get your point, and I think I know where the conflict comes from: you are a bit more strict about the interpretation of `_is_valid_input(...) == false`. IMHO it just indicates that the input is not valid, which might not necessarily mean, that this is an you need to catch. It also could be used as a selector/mask-generator for elements of a vector, where `_is_valid_input(...) == false` just elimitates the *entry* but not throwing an error. Therefore, `_is_valid_input` should be the *faster* version a user might implement. 

But you are right, of course, if we want to be more strict about the meaning of *input validation*, then `_is_valid_input(...) == false` should be interreted as wrong input and therefore throw an `Exception`, which needs to be handled from the layer above. I will adjust the code accordingly, and send another post to this thread, when its finished.

tjungni (dismissed):
From my side everything is resolved.

src/interfaces/setup_interface.jl
AntonReinhard: Please still fix the spaces in the error string here

AntonReinhard (approved):

src/interfaces/setup_interface.jl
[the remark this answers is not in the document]
szabo137: Applied the suggested changes in c3d4456.

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,226 @@
...
+
+    _is_valid_input(stp::AbstractComputationSetup, input::Any)
+
+Interface function, which returns true if the constraints of the `input` associated with the quantity of `stp` are met.
+This function is called to validate the input of [`compute`](@ref) before calling [`_compute`](@ref).
+
+!!! note "Default implementation"
+    
+    Since no input validation is equivalent to every input being valid, this function returns `true` by default. 
+    This behavior can be overwritten if actual validation is necessary.
```
AntonReinhard, line 121: This is not really true anymore now. Only if _assert_valid_input is not overwritten is this true.

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,226 @@
...
+"""
+function _compute end
+
+"""
+
+    compute(stp::AbstractComputationSetup, input::Any)
+
+Return the value of the quantity associated with `stp` for a given `input`. 
+In addition to the actual call of the associated unsafe version [`_compute`](@ref),
+input validation ([`_assert_valid_input`]) and post processing
```
AntonReinhard, line 176:
```suggestion
input validation ([`_assert_valid_input`](@ref)) and post processing 
```

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,226 @@
...
+    The generic fallback uses the boolian value returned by `_is_valid_input(stp,input)` to check for validity, 
+    i.e. if the returned value is `true` the assert does not trigger and nothing happens, but if the returned value is `false`, 
+    an error with a generic error message will be thrown (see  [`_is_valid_input`](@ref) for details). 
+    To customize the error message, or use a custom assertion mechanism, just add your own implementation of
+
+    ```Julia
+    _assert_valid_input(stp::YourCustomSetup,input)
+    ```
+    which should throw an [`InvalidInputError`](@ref) if the input is invalid.
+    Despite the presence of a custom `_assert_valid_input`, it is highly recommended to also implement `_is_valid_input` for `CustomSetup`, because it might be used outside of the assert function.
```
AntonReinhard, line 104: This is also not true anymore.
(also boolian => boolean typo ;))

src/interfaces/setup_interface.jl
```diff
@@ -0,0 +1,226 @@
...
+
+"""
+    
+    _compute(stp::AbstractComputationSetup, input::Any)
+
+Interface function that returns the value of the associated quantity evaluated on `input`, which can be anything the associated quantity is defined to be feasible for.
+
+!!! note "unsafe implementation"
+
+    This function must be implemented for any subtype of [`AbstractComputationSetup`]. It should not do any input validation or post processing (see [`_is_valid_input`](@ref) and [`_post_processing`](@ref)), as those two are performed while calling
```
AntonReinhard, line 164: Should this reference `_assert_valid_input` instead of `_is_valid_input`, since that is the one the user of the interface should implement (first)?
```suggestion
    This function must be implemented for any subtype of [`AbstractComputationSetup`]. It should not do any input validation or post processing (see [`_assert_valid_input`](@ref) and [`_post_processing`](@ref)), as those two are performed while calling 
```

src/interfaces/setup_interface.jl
```diff
@@ -92,51 +92,19 @@ Interface function, which asserts that the given `input` is valid, and throws an
 
 !!! note "default implementation"
 
-    The generic fallback uses the boolian value returned by `_is_valid_input(stp,input)` to check for validity, 
-    i.e. if the returned value is `true` the assert does not trigger and nothing happens, but if the returned value is `false`, 
-    an error with a generic error message will be thrown (see  [`_is_valid_input`](@ref) for details). 
-    To customize the error message, or use a custom assertion mechanism, just add your own implementation of
+    By default, every input is assumed to be valid. Therefore, this functions does nothing.
```
AntonReinhard, line 95:
```suggestion
    By default, every input is assumed to be valid. Therefore, this function does nothing. 
```

src/interfaces/setup_interface.jl
```diff
@@ -30,7 +30,7 @@ quantity depends on is kept constant.
     _assert_valid_input(stp::AbstractComputationSetup, input)
     ```
     
-    which should throw and an exception subtyped from [`AbstractInvalidInputException`](@ref) if the `input` is not valid for the computation of the associated quantity (see [`_is_valid_input`](@ref) and [`_assert_valid_input`](@ref) for more details). 
+    which should throw and an exception subtyped from [`AbstractInvalidInputException`](@ref) if the `input` is not valid for the computation of the associated quantity ([`_assert_valid_input`](@ref) for more details).
```
AntonReinhard, line 33:
```suggestion
    which should throw and an exception subtyped from [`AbstractInvalidInputException`](@ref) if the `input` is not valid for the computation of the associated quantity (see [`_assert_valid_input`](@ref) for more details). 
```

src/interfaces/setup_interface.jl
```diff
@@ -92,51 +92,19 @@ Interface function, which asserts that the given `input` is valid, and throws an
...
-    To customize the error message, or use a custom assertion mechanism, just add your own implementation of
+    By default, every input is assumed to be valid. Therefore, this functions does nothing. 
+    To customize this behavior add your own implementation of
 
     ```Julia
     _assert_valid_input(stp::YourCustomSetup,input)
     ```
-    which should throw an [`InvalidInputError`](@ref) if the input is invalid.
-    Despite the presence of a custom `_assert_valid_input`, it is highly recommended to also implement `_is_valid_input` for `CustomSetup`, because it might be used outside of the assert function.
+    which should throw an exception, which is subtype of [`AbstractInvalidInputError`](@ref). One may also use the concrete implementation [`InvalidInputError`](@ref) if the input is invalid, instead of writing a custom exception type.
```
AntonReinhard, line 101:
```suggestion
    which should throw an exception, which is a subtype of [`AbstractInvalidInputError`](@ref). One may also use the concrete implementation [`InvalidInputError`](@ref) if the input is invalid instead of writing a custom exception type.
```

src/interfaces/setup_interface.jl
```diff
@@ -161,7 +129,7 @@ Interface function that returns the value of the associated quantity evaluated o
 
 !!! note "unsafe implementation"
 
-    This function must be implemented for any subtype of [`AbstractComputationSetup`]. It should not do any input validation or post processing (see [`_is_valid_input`](@ref) and [`_post_processing`](@ref)), as those two are performed while calling 
+    This function must be implemented for any subtype of [`AbstractComputationSetup`]. It should not do any input validation or post processing (see [`_assert_valid_input`](@ref) and [`_post_processing`](@ref)), as those two are performed while calling
```
AntonReinhard, line 132:
```suggestion
    This function must be implemented for any subtype of [`AbstractComputationSetup`](@ref). It should not do any input validation or post processing (see [`_assert_valid_input`](@ref) and [`_post_processing`](@ref)), as those two are performed while calling 
```

src/interfaces/setup_interface.jl
```diff
@@ -92,51 +92,19 @@ Interface function, which asserts that the given `input` is valid, and throws an
 
 !!! note "default implementation"
 
-    The generic fallback uses the boolian value returned by `_is_valid_input(stp,input)` to check for validity, 
-    i.e. if the returned value is `true` the assert does not trigger and nothing happens, but if the returned value is `false`, 
-    an error with a generic error message will be thrown (see  [`_is_valid_input`](@ref) for details). 
-    To customize the error message, or use a custom assertion mechanism, just add your own implementation of
+    By default, every input is assumed to be valid. Therefore, this functions does nothing. 
+    To customize this behavior add your own implementation of
```
AntonReinhard, line 96:
```suggestion
    To customize this behavior, add your own implementation of
```

AntonReinhard (dismissed):

AntonReinhard (approved):

PR #18, opened 2023-10-05
Add Differential Cross Section interface, Compton and PerturbativeQED Model

AntonReinhard:
CI_UNIT_PKG_URL_QEDbase: https://github.com/AntonReinhard/QEDbase.jl#1735e72

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,10 +4,8 @@ authors = ["Uwe Hernandez Acosta <u.hernandez@hzdr.de>", "Simeon Ehrig", "Klaus 
 version = "0.1.0"
 
 [deps]
-DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 QEDbase = "10e22c08-3ccb-4172-bfcf-7d7aa3d04d93"
 
 [compat]
-DocStringExtensions = "0.9"
 QEDbase = "0.1"
 julia = "1.6"
--- a/src/QEDprocesses.jl
+++ b/src/QEDprocesses.jl
@@ -7,7 +7,14 @@ export AbstractModelDefinition, fundamental_interaction_type
 export AbstractProcessDefinition, incoming_particles, outgoing_particles
 export in_phasespace_dimension, out_phasespace_dimension
 export number_incoming_particles, number_outgoing_particles
-export differential_cross_section, total_cross_section
+
+# probabilities
+export differential_probability, unsafe_differential_probability
+export total_probability
+
+# differential cross section
+export differential_cross_section, unsafe_differential_cross_section
+export total_cross_section
 
 # Abstract setup interface
 export AbstractComputationSetup, InvalidInputError, compute
@@ -16,7 +23,11 @@ export AbstractProcessSetup, scattering_process, physical_model
 # propagator
 export propagator
 
-using DocStringExtensions
+# phase space 
+export AbstractCoordinateSystem, SphericalCoordinateSystem
+export AbstractFrameOfReference, CenterOfMomentumFrame, ElectronRestFrame
+export AbstractPhasespaceDefinition, PhasespaceDefinition
+
 using QEDbase
 
 include("utils.jl")
@@ -24,4 +35,7 @@ include("interfaces/model_interface.jl")
 include("interfaces/process_interface.jl")
 include("interfaces/setup_interface.jl")
 include("propagators.jl")
+include("phase_spaces.jl")
+include("probabilities.jl")
+include("cross_sections.jl")
 end
--- a/src/cross_sections.jl
+++ b/src/cross_sections.jl
@@ -1,0 +1,304 @@
+########################
+# differential cross sections and probabilities.
+#
+# This file contains default implementations for differential and total cross
+# sections based on the scattering process interface
+########################
+
+############
+#
+# differential cross sections
+#
+############
+
+# differential cross sections without energy momentum conservation check
+# single in phase space point/ single out phase space point
+function _unsafe_differential_cross_section(
+    proc::AbstractProcessDefinition,
+    model::AbstractModelDefinition,
+    in_phase_space_def::AbstractPhasespaceDefinition,
+    in_phase_space::AbstractVector{T},
+    out_phase_space_def::AbstractPhasespaceDefinition,
+    out_phase_space::AbstractVector{T},
+) where {T<:QEDbase.AbstractFourMomentum}
+    I = 1 / (4 * _incident_flux(proc, model, in_phase_space))
+
+    return I * _unsafe_differential_probability(
+        proc,
+        model,
+        in_phase_space_def,
+        in_phase_space,
+        out_phase_space_def,
+        out_phase_space,
+    )
+end
+
+# differential cross sections without energy momentum conservation check
+# single in phase space point/ several out phase space points
+function _unsafe_differential_cross_section(
+    proc::AbstractProcessDefinition,
+    model::AbstractModelDefinition,
+    in_phase_space_def::AbstractPhasespaceDefinition,
+    in_phase_space::AbstractVector{T},
+    out_phase_space_def::AbstractPhasespaceDefinition,
+    out_phase_space::AbstractMatrix{T},
+) where {T<:QEDbase.AbstractFourMomentum}
+    res = Vector{eltype(T)}(undef, size(out_phase_space, 2))
+    for i in 1:size(out_phase_space, 2)
+        res[i] = _unsafe_differential_cross_section(
+            proc,
+            model,
+            in_phase_space_def,
+            in_phase_space,
+            out_phase_space_def,
+            view(out_phase_space, :, i),
+        )
+    end
+    return res
+end
+
+# differential cross sections without energy momentum conservation check
+# several in phase space points/ one or several out phase space points
+function _unsafe_differential_cross_section(
+    proc::AbstractProcessDefinition,
+    model::AbstractModelDefinition,
+    in_phase_space_def::AbstractPhasespaceDefinition,
+    in_phase_space::AbstractMatrix{T},
+    out_phase_space_def::AbstractPhasespaceDefinition,
+    out_phase_space::AbstractVecOrMat{T},
+) where {T<:QEDbase.AbstractFourMomentum}
+    res = Matrix{eltype(T)}(undef, size(in_phase_space, 2), size(out_phase_space, 2))
+    for i in 1:size(in_phase_space, 2)
+        res[i, :] .= _unsafe_differential_cross_section(
+            proc,
+            model,
+            in_phase_space_def,
+            view(in_phase_space, :, i),
+            out_phase_space_def,
+            out_phase_space,
+        )
+    end
+    return res
+end
+
+"""
+
+    function unsafe_differential_cross_section(
+        proc::AbstractProcessDefinition,
+        model::AbstractModelDefinition,
+        in_phase_space_def::AbstractPhasespaceDefinition,
+        in_phase_space::AbstractVecOrMat{T},
+        out_phase_space_def::AbstractPhasespaceDefinition,
+        out_phase_space::AbstractVecOrMat{T},
+    ) where {T<:QEDbase.AbstractFourMomentum}
+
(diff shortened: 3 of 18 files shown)


AntonReinhard:
Currently fails because it depends on https://github.com/QEDjl-project/QEDbase.jl/pull/25

src/differential_cross_section.jl
```diff
@@ -0,0 +1,93 @@
...
+        InvalidInputError(
+            "In phase space dimension is inconsistent with input size ($(size(dcs.in_phase_space)) versus $(number_incoming_particles(dcs.process)))",
+        ),
+    )
+    length(out_phase_space) == number_outgoing_particles(dcs.process) || throw(
+        InvalidInputError(
+            "Out phase space dimension is inconsistent with input size ($(size(out_phase_space, 1)) versus $(number_outgoing_particles(dcs.process)))",
+        ),
+    )
+    _assert_valid_input(dcs.process, dcs.model, dcs.in_phase_space, out_phase_space)
```
szabo137, line 84: `_assert_valid_dcs_input`?

src/differential_cross_section.jl
```diff
@@ -0,0 +1,93 @@
...
+function _assert_valid_dcs_input end
+
+"""
+    _differential_cross_section(model::AbstractModelDefinition, process::AbstractScatteringProcess, phaseSpace::AbstractArray{NumericType}, input::AbstractArray{NumericType}) where {NumericType}
+
+Interface function that must be implemented for valid model-process combinations. This function will be dispatched to when [`_compute`](@ref) is called on a [`DifferentialCrossSection`](@ref), which in turn is called in [`compute`](@ref).
+
+For details on what it should do, see [`_compute`](@ref).
+"""
+function _differential_cross_section end
```
szabo137, line 46: Is this really necessary? The interface is actually defined in `interfaces/process_interface.jl`.

src/differential_cross_section.jl
```diff
@@ -0,0 +1,93 @@
...
+
+"""
+    _post_process(model::AbstractModelDefinition, process::AbstractScatteringProcess, phaseSpace::AbstractArray{NumericType}, input::AbstractArray{NumericType}, result::Any) where {NumericType}
+
+Interface function that must be implemented for valid model-process combinations. This function will be dispatched to when [`_post_process`](@ref) is called on a [`DifferentialCrossSection`](@ref), which in turn is called in [`compute`](@ref).
+
+For details on what it should do, see [`_post_process`](@ref).
+"""
+function _post_process end
```
szabo137, line 55: I think it is highly misleading if an interface function defined for setups is dispatched on non-setup types. I understand the meaning, but suggest renaming it.

src/differential_cross_section.jl
```diff
@@ -0,0 +1,93 @@
...
+    isa(dcs.in_phase_space, AbstractArray{NumericType}) || throw(
+        InvalidInputError(
+            "Input type is '$(typeof(out_phase_space))' but '$(typeof(dcs.in_phase_space))' was expected",
+        ),
+    )
+    length(dcs.in_phase_space) == number_incoming_particles(dcs.process) || throw(
+        InvalidInputError(
+            "In phase space dimension is inconsistent with input size ($(size(dcs.in_phase_space)) versus $(number_incoming_particles(dcs.process)))",
+        ),
+    )
```
szabo137, line 78: Input validation for `in_phasespace` should be done in the constructor of the setup, i.e. in `DifferentialCrossSection`. (This is the purpose of the setup in the first place).

test/compton.jl
```diff
@@ -0,0 +1,61 @@
...
+    @testset "default" begin
+        proc = Compton()
+        @test QEDprocesses._spin_or_pol(proc, Photon(), Incoming()) == AllPol()
+        @test QEDprocesses._spin_or_pol(proc, Electron(), Incoming()) == AllSpin()
+        @test QEDprocesses._spin_or_pol(proc, Photon(), Outgoing()) == AllPol()
+        @test QEDprocesses._spin_or_pol(proc, Electron(), Outgoing()) == AllSpin()
+    end
+
+    @testset "non-default" begin
+        proc = Compton((PolX(), PolY()), (SpinUp(), SpinDown()))
```
szabo137, line 15: maybe looping over all definite spin and polarisation types.

test/compton.jl
```diff
@@ -0,0 +1,61 @@
...
+        @test QEDprocesses._spin_or_pol(proc, Photon(), Incoming()) == PolX()
+        @test QEDprocesses._spin_or_pol(proc, Electron(), Incoming()) == SpinUp()
+        @test QEDprocesses._spin_or_pol(proc, Photon(), Outgoing()) == PolY()
+        @test QEDprocesses._spin_or_pol(proc, Electron(), Outgoing()) == SpinDown()
+    end
+end
+
+@testset "scattering process interface" begin
```
szabo137, line 23: This should also be tested for all combinations of spin and polarisation types.

test/compton.jl
```diff
@@ -0,0 +1,61 @@
...
+            valid,
+        )
+        @test_throws "incoming" differential_cross_section(proc, model, invalid, valid)
+        @test_throws DimensionMismatch differential_cross_section(
+            proc,
+            model,
+            valid,
+            invalid,
+        )
+        @test_throws "outgoing" differential_cross_section(proc, model, valid, invalid)
```
szabo137, line 59: The thrown Exception is a string?

szabo137 (commented):
First comments on the draft.

SimeonEhrig:
> Currently fails because it depends on [QEDjl-project/QEDbase.jl#25](https://github.com/QEDjl-project/QEDbase.jl/pull/25)

You can run the unit tests with a development branch as dependency: https://github.com/QEDjl-project/QED.jl/blob/dev/docs/src/ci.md#unit-tests-for-ci-users

src/compton/differential_cross_section.jl
```diff
@@ -0,0 +1,108 @@
...
+    return nothing
+end
+
+function _differential_cross_section(
+    process::Compton,
+    model::PerturbativeQED,
+    in_phase_space::AbstractVector{NumericType},
+    out_phase_space::AbstractVector{NumericType},
+)::Float64 where {NumericType<:QEDbase.AbstractFourMomentum}
+    if (!isapprox(sum(in_phase_space), sum(out_phase_space); rtol = sqrt(eps())))
```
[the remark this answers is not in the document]
AntonReinhard: I thought the idea was that `_differential_cross_section` is inherently unsafe and any assertions happen in `_assert_valid_input`, which is where the on-shell-ness is tested currently.

test/compton.jl
AntonReinhard: No. The thrown exception is a `DimensionMismatch`. Giving a string here matches the string against the exception's string. The type of exception is checked a bit higher. I did not see a way to check both the type and do string matching in one test so I just made two separate ones.

test/compton.jl
AntonReinhard: Aren't these all of them? Or is there a way to directly iterate over all concrete subtypes of a given abstract type?

AntonReinhard: Nevermind I understand what you meant :sweat_smile:

src/differential_cross_section.jl
AntonReinhard: You're right. I renamed it to `_post_process_dcs`, is that fine?

src/differential_cross_section.jl
AntonReinhard: You're completely right, I removed it.

src/compton/differential_cross_section.jl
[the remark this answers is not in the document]
szabo137: This is correct, but the check for on-shell momenta is not *input validation* but a physics, i.e. the differential cross section is zero, if any of the external momenta does not obey `p*p=mass(...)^2`, similar to the energy-momentum conservation. The word *unsafe* means just an implementation of this function, which assumes, that the input momenta are all on-shell and conserve momentum and energy. The reason for this is, that it should be possible, to check this somewhere else, and therefore not again in the differential cross-section. For example, RAMBO produces momenta, which are automatically on-shell and energy-momentum conserving, which means these properties must not be checked here again.

test/compton.jl
szabo137: Ok, understood, then it is fine, of course.

AntonReinhard:
The docs building step currently fails because referenced types/functions have moved to QEDbase now, I'm not sure how to proceed with that. I don't want to delete the references or put the links to the QEDbase docs in manually.

Maybe MultiDocumenter would handle this?

SimeonEhrig:
I don't think it solves the problem. `MultiDocumenter` compiles only the rendered HTML documentation of different packages.

I found this issue: https://github.com/JuliaDocs/Documenter.jl/issues/688
So it is not official documented. 

Maybe we can use as workaround markdown links `[foo()](doc.com/api#foo)`, if we have a API reference page.

src/compton/differential_cross_section.jl
```diff
@@ -0,0 +1,202 @@
...
+        base_state(
+            Electron(),
+            Incoming(),
+            electron_in,
+            _spin_or_pol(process, Electron(), Incoming()),
+        ),
+    )
+
+    # average over incoming polarizations/spins, but sum over outgoing pols/spins
+    normalization = 1.0 / (length(photon_in_bstate) * length(electron_in_bstate))
```
AntonReinhard, line 56: @szabo137 How should we do this?
Some anonymous or _ function that just maps from AbstractPol/AbstractSpin to 1 or 2?

AntonReinhard:
> I don't think it solves the problem. `MultiDocumenter` compiles only the rendered HTML documentation of different packages.
> 
> I found this issue: [JuliaDocs/Documenter.jl#688](https://github.com/JuliaDocs/Documenter.jl/issues/688) So it is not official documented.
> 
> Maybe we can use as workaround markdown links `[foo()](doc.com/api#foo)`, if we have a API reference page.

This would work, but for example `AbstractParticle` alone is referenced 8 times already, so I'd have to put that link into lots of places... I guess there's not really a better way right now though. Other than not referencing at all.

src/compton/differential_cross_section.jl
szabo137: I think this should be another interface function for scattering processes, e.g. `averaging_norm(proc::AbstractScatteringProcess)`. For the specific implementation, I think a map of the `AbstractSpin/AbstractPol` to `1` or `2` seems convenient. 

Nevertheless, I am currently working on a refactoring of the `differential_cross_section` function, which includes all internal functions like `incident_flux` or `matrix_element` as interface functions on `AbstractScatteirngProcess`. Maybe we should discuss the internals of the current implementation of `differential_cross_section` in this context.

szabo137:
I have another crossref for the docstring issue: https://github.com/JuliaDocs/Documenter.jl/issues/1343

Another workaround would be, to include the docstring of the at_ref function/type to the documentation of `QEDprocesses`, e.g. in the `api.md` page. However, this is a bit hacky, because it means to include `QEDbase` in the `makedocs` of `QEDprocesses` (I can provide a showcase or PR if needed). In this case, at_ref works again without specifying the URL, but it seems to be a very ugly solution.

src/compton/differential_cross_section.jl
AntonReinhard: Ah that makes sense. Do you want to make a PR into this branch or how do you plan to do it?

szabo137: I think `differential_cross_section` can be disentangled from this one here. Therefore I suggest making a separate PR from `dev`, and after merging, we can update this one here.

szabo137: I started the refactoring of `differential_cross_section` here: https://github.com/QEDjl-project/QEDprocesses.jl/pull/38 

This changes the scattering process interface from `differential_cross_section` to more fine granular functions, i.e. `incident_flux`, `averaging_norm`, `matrix_element`, and `phase_space_factor`. The actual `differential_cross_section` is downgraded from an interface function to a default implementation using the new interface. Currently, it's a draft-PR because it is not finished yet and the tests might be broken.

AntonReinhard:
I'm sorry, I somehow managed to kill this PR by accident. I saved the commits but I might have to open a new PR.

PR #19, opened 2023-10-10
add document build and deploy jobs

SimeonEhrig:
solves #10 

Copy of: https://github.com/QEDjl-project/QEDevents.jl/pull/10

### DIFF
--- a/.github/workflows/BuildDelopyDoc.yml
+++ b/.github/workflows/BuildDelopyDoc.yml
@@ -1,0 +1,31 @@
+name: Build and Deploy Documentation
+
+on:
+  push:
+    branches:
+      - master
+      - dev
+    tags: '*'
+  pull_request:
+
+jobs:
+  build:
+    permissions:
+      contents: write
+      statuses: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1.9'
+      - name: Add custom registry 
+        run: |
+          julia --project=docs/ -e 'import Pkg; Pkg.Registry.add(Pkg.RegistrySpec(url="https://github.com/QEDjl-project/registry.git"));'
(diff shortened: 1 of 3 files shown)


SimeonEhrig:
> I will look into this. I assume some API change in `Documenter.jl`.

Looks like this line of code causes the problem: https://github.com/QEDjl-project/QEDprocesses.jl/blob/0074aa02285d5ef39e63f5a204e02ac4b5b1f6d8/src/interfaces/setup_interface.jl#L101

SimeonEhrig:
> > I will look into this. I assume some API change in `Documenter.jl`.
> 
> Looks like this line of code causes the problem:
> 
> https://github.com/QEDjl-project/QEDprocesses.jl/blob/0074aa02285d5ef39e63f5a204e02ac4b5b1f6d8/src/interfaces/setup_interface.jl#L101

@szabo137 Looks like, the documentation references an error type, which does not exist. I didn't found anything in the QED projects and also not in the official documentation.

AntonReinhard:
It should be `AbstractInvalidInputException`, that's what is defined. But it is a little weird that one is `SomethingException` and the other is `SomethingError`, so we should probably use one of the two consistently everywhere. I think I would prefer `Exception`,  `Error` is less specific.

SimeonEhrig:
> It should be `AbstractInvalidInputException`, that's what is defined. But it is a little weird that one is `SomethingException` and the other is `SomethingError`, so we should probably use one of the two consistently everywhere. I think I would prefer `Exception`, `Error` is less specific.
> 
> So I'd suggest renaming the `InvalidInputError` to `InvalidInputException`, too.

Yes, you are right. We are a little bit inconsistent :sweat_smile: 

https://github.com/QEDjl-project/QEDprocesses.jl/blob/0074aa02285d5ef39e63f5a204e02ac4b5b1f6d8/src/interfaces/setup_interface.jl#L79-L83

Nevertheless, I would fix the error in the documentation now and we do the possible renaming in an extra PR.

szabo137 (approved):
Nothing to bother, merging!

PR #28, opened 2023-10-18
Add Propagators

szabo137:
This PR adds a simple implementation of fermion and boson propagators. 

This should be merged *before* #18 because the latter will use the functionality. 

TODOS: 

- [x] replace dev-version of `QEDbase` with release `v0.1.5` (cf. https://github.com/QEDjl-project/QEDbase.jl/issues/39)

### DIFF
--- a/src/propagators.jl
+++ b/src/propagators.jl
@@ -36,25 +36,26 @@ Return the propagator of a particle for a given four-momentum. If `mass` is pass
 function propagator end
 
 function _scalar_propagator(K::QEDbase.AbstractFourMomentum, mass::Real)
-    one(mass) / (K * K - mass^2)
+    return one(mass) / (K * K - mass^2)
 end
 
 function _scalar_propagator(K::QEDbase.AbstractFourMomentum)
-    one(getT(K)) / (K * K)
+    return one(getT(K)) / (K * K)
 end
 
-
 function _fermion_propagator(P::QEDbase.AbstractFourMomentum, mass::Real)
-    (slashed(P) + mass * one(DiracMatrix)) * _scalar_propagator(P, mass)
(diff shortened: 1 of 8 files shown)


szabo137:
@AntonReinhard This should be ready for review now. Could you please have a look on this?

src/propagators.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+function _fermion_propagator(P::QEDbase.AbstractFourMomentum)
+    return (slashed(P)) * _scalar_propagator(P)
+end
+
+function propagator(particle_type::BosonLike, K::QEDbase.AbstractFourMomentum)
+    return _scalar_propagator(K, mass(particle_type))
+end
+propagator(particle_type::Photon, K::QEDbase.AbstractFourMomentum) = _scalar_propagator(K)
```
AntonReinhard, line 57: Can we make this into a whole `function ... end` too just for the niceness?

test/propagators.jl
```diff
@@ -0,0 +1,26 @@
+using Random
+using QEDbase
+using QEDprocesses
+
+RNG = MersenneTwister(137137)
+ATOL = 0.0
```
AntonReinhard, line 6: atol defaults to zero anyways

AntonReinhard (changes requested):

test/propagators.jl
szabo137: This was made explicitly to not rely on defaults of `isapprox`, and it is better to read.

src/propagators.jl
szabo137: Niceness reestablished in 77c5494

AntonReinhard: Thank you ^^

AntonReinhard (approved):

PR #29, opened 2023-10-19
add missing .JuliaFormatter.toml and format all files

SimeonEhrig:


szabo137 (approved):
Thanks for adding this!

No additional comments.

PR #30, opened 2023-10-19
run compat helper only on upstream repository

SimeonEhrig:
This if-condition avoids, that the compat helper is executed in forks.

copy of: https://github.com/QEDjl-project/QEDfields.jl/pull/19

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -8,6 +8,7 @@ permissions:
   pull-requests: write
 jobs:
   CompatHelper:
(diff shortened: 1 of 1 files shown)


szabo137 (approved):
Thanks for this fix! No concerns from my side, merging!

SimeonEhrig:
I pushed the current dev branch to my fork. I will check the compat helper jobs of upstream and fork tomorrow. If everything is like expected, I will also open the PR's for the other packages.

SimeonEhrig:
Works like expected. I will open the PR's in the other projects.

PR #39, opened 2023-11-30
Coordinate-based evaluation of differential cross-sections and probabilities

szabo137:
# Problem description

Currently, differential cross-sections and probabilities can only computed on given momenta. However, in some cases, one wants to calculate those quantities directly on given phase-space parameterizations, i.e. independent coordinates.

# Suggested solution

One might use multiple dispatch on the element type of the incoming and outgoing phase-space of `differential_cross_section` and `probability` to determine if coordinates or momenta are passed in. Furthermore, it seems sufficient to use the following implementation of `differential_cross_section` as the actual interface function

The coordinate-based implementation of the differential cross-section should then fall back onto the momentum-based implementation by a phase-space generator, e.g.

```Julia
generate_momenta(
     in_phasespace_def,
     in_phasespace::AbstractVecOrMat{T},
     out_phasespace_def,
     out_phasespace::AbstractVecOrMat{T}
     ) where {T<:Real}
```
which returns the respective vector or matrix of `SFourMomentum`. Using that, the default implementation of the `differential_cross_section` reads

```Julia
function _differential_cross_section(
    proc, model,
    in_phasespace_def,
    in_phasespace::AbstractVecOrMat{T},
    out_phasespace_def,
    out_phasespace:: AbstractVecOrMat{T}
    ) where {T<:Real}
    in_moms, out_moms = generate_momenta(proc, model, in_phasespace_def, out_phasespace_def, in_phasespace, out_phasespace)
    return _differential_cross_section(proc, model, in_phasespace_def, out_phasespace_def, in_moms, out_moms)
end
```
which works generically. 

For the evaluation of cross sections and probabilities on sets of phase space points, the 
implementation is the same for momenta and coordinates. Therefore, this PR introduces a general phase space element type: `AbstractPhaseSpacePoint = Union{Real, AbstractFourMomentum}`. 

## TODO

- [x] Write docs for added versions of the exported functions.
- [ ] Write an issue on the phase space interface including the momentum generation above.


szabo137:
@steindev @tjungni If I did not forget something, this PR is ready from my side. I would appreciate some comments from your side. 🙆‍♂️

src/phase_spaces.jl
```diff
@@ -1,8 +1,10 @@
 #####################
 # phase spaces  
 #
-# This file contains a collection of types and functions to handle phase spaces
+# This file contains a colection of types and functions to handle phase spaces
```
tjungni, line 4:
```suggestion
# This file contains a collection of types and functions to handle phase spaces
```

src/phase_spaces.jl
```diff
@@ -1,8 +1,10 @@
...
 # for scattering processes.
+#
+# TODO: ship to interfaces!
```
tjungni, line 7: Just a reminder that there is a todo here. Not sure if it is part of this PR.
See also `in_phase_space_dimension` and `out_phase_space_dimension` in process_interface.jl

src/probabilities.jl
```diff
@@ -124,8 +148,51 @@ function unsafe_differential_probability(
...
+        model,
+        in_phase_space_def,
+        in_phase_space,
+        out_phase_space_def,
+        out_phase_space,
+    )
+end
+
 # differential probability with energy momentum conservation check
 # one in phase space points/ one out phase space point
```
tjungni, line 194:
```suggestion
# one in phase space point/ one out phase space point
```

src/probabilities.jl
```diff
@@ -155,6 +222,30 @@ function _differential_probability(
     )
 end
 
+# differential probability with energy momentum conservation check
+# one in phase space points/ one out phase space point
```
tjungni, line 226:
```suggestion
# one in phase space point/ one out phase space point
```

test/test_implementation/utils.jl
```diff
@@ -3,3 +3,37 @@
...
+    for i in 1:size(ps_moms, 2)
+        res[:, i] .= _unroll_moms(view(ps_moms, :, i))
+    end
+    return res
+end
+
+flat_components(moms) = _unroll_moms(moms)
+
+# collect components of four-momenta from a vector of coordinates
+function __furl_moms(ps_coords::AbstractVector{T}) where {T<:Real}
```
tjungni, line 23:
```suggestion
function _furl_moms(ps_coords::AbstractVector{T}) where {T<:Real}
```

tjungni (commented):
Looks great! I don't have much to add

src/phase_spaces.jl
szabo137: Thanks for pointing out, this will be part of another PR.

src/interfaces/process_interface.jl
```diff
@@ -200,6 +200,24 @@ Return the number of outgoing particles of a given process.
...
+function in_phase_space_dimension end
+
+"""
+    out_phase_space_dimension(
+        proc::AbstractProcessDefinition,
+        model::AbstractModelDefinition,
+        )
+TBW
+"""
+function out_phase_space_dimension end
```
AntonReinhard, line 219: Is there an issue for these that could be referenced?

src/cross_sections.jl
```diff
@@ -302,3 +443,28 @@ function total_cross_section(
...
+    proc::AbstractProcessDefinition,
+    model::AbstractModelDefinition,
+    in_phase_space_def::AbstractPhasespaceDefinition,
+    in_phase_space::AbstractVecOrMat{T},
+) where {T<:Real}
+    size(in_phase_space, 1) == in_phase_space_dimension(proc, model) || throw(
+        DimensionMismatch(
+            "The dimension of the in-phase-space <$(in_phase_space_dimension(proc,model))> is inconsistent with input size <$(size(in_phase_space,1))>",
+        ),
+    )
```
AntonReinhard, line 467: These lines appear many times basically unchanged. It might make sense to create something like a `_check_phase_space_dimension()` that does this.

AntonReinhard (commented):
This seems like a lot of code duplication for not too much gain. If I understand correctly, the idea is to allow a new, different type of representation of the FourMomenta as input. So why not make the existing functions generically accept any `AbstractPhasespaceElement` and then a function (or multiple functions) that convert an `AbstractPhasespaceElement` into a specific representation? 
The interface of this function could even completely generically allow to specify a target type, so if more types might be added later, the only thing that needs to be added are the relevant conversion function implementations.
This function would then be the identity function if it's already the correct representation, and otherwise do whatever it needs to do. This would be the `_generate_incoming_momenta()` I believe.
Also, it might make sense to have this function only work on a single element and then broadcast where necessary, seems more flexible.

src/interfaces/process_interface.jl
szabo137: Opened an issue on that: #48

src/cross_sections.jl
szabo137: Thanks for the suggestion, added it with 51b583b.

szabo137:
@AntonReinhard could you please check is something is left over here? If not, we could proceed with this one 😊

AntonReinhard (approved):

PR #40, opened 2024-02-22
perturbative Compton

szabo137:
This PR adds one photon Compton scattering described in perturbative QED. 

This includes the following variants
* diff. probability and cross section on momenta
* diff. probability and cross section on coordinates (electron restframe and spherical coordinates)

All implementations are done for arbitrary combinations of spins and polarizations.

## Todos

- [x] discuss where the spins/polarizations are stored: proc or phase space point (be aware of dispatch)
- [x] add total cross section
- [x] add unit test for total cross section
- [x] add unit tests for phase space point input


AntonReinhard:
What's the general state of this now?
Trying to rebase to dev seems to be a nightmare, but merging dev back into this branch should be pretty easy. Other than that, some things need to be adjusted now to the `PhaseSpacePoints` and `ParticleStateful` types I believe.

szabo137:
> What's the general state of this now? Trying to rebase to dev seems to be a nightmare, but merging dev back into this branch should be pretty easy. Other than that, some things need to be adjusted now to the `PhaseSpacePoints` and `ParticleStateful` types I believe.

I followed your advice and merged `dev` into this one. Now this should be up-to-date and ready for discussion. There are some open questions we could think about first (see the description box above).

src/models/perturbative_qed.jl
```diff
@@ -0,0 +1,23 @@
...
+
+"""
+    in_phase_space_dimension(proc::AbstractProcessDefinition, ::PerturbativeQED)
+
+
+Return the number of degrees of freedom to determine the incoming phase space for processes in PerturbativeQED. 
+
+!!! note "Convention"
+
+    The current implementation only supports the case, where two of the incoming particles collide heads-on.
```
AntonReinhard, line 14:
```suggestion
    The current implementation only supports the case where two of the incoming particles collide head-on. 
```

src/patch_QEDbase.jl
```diff
@@ -9,3 +9,8 @@ Broadcast.broadcastable(dir::Incoming) = Ref(dir)
 Broadcast.broadcastable(dir::Outgoing) = Ref(dir)
 Broadcast.broadcastable(part::AbstractParticleType) = Ref(part)
 Broadcast.broadcastable(spin_or_pol::AbstractSpinOrPolarization) = Ref(spin_or_pol)
+
+number_of_spin_pol(::AbstractDefinitePolarization) = 1
+number_of_spin_pol(::AbstractDefiniteSpin) = 1
+number_of_spin_pol(::AbstractIndefinitePolarization) = 2
+number_of_spin_pol(::AbstractIndefiniteSpin) = 2
```
AntonReinhard, line 16: Open/link additional issue in QEDbase for this?

src/processes/one_photon_compton/one_photon_compton.jl
```diff
@@ -0,0 +1,12 @@
+#################
+# The one-photon Compton process
+#
+# This file contains the implementation of the abstract process setup for
+# Compton
+##################
+
+export Compton
```
AntonReinhard, line 8: Already exported in QEDprocesses.jl

src/processes/one_photon_compton/perturbative/cross_section.jl
```diff
@@ -0,0 +1,173 @@
...
+end
+
+function _is_in_phasespace(
+    proc::Compton,
+    model::PerturbativeQED,
+    in_ps_def::AbstractPhasespaceDefinition,
+    in_ps::AbstractVector{T},
+    out_ps::AbstractVector{T},
+) where {T<:QEDbase.AbstractFourMomentum}
+    return (!isapprox(sum(in_ps), sum(out_ps))) ? false : _all_onshell(proc, in_ps, out_ps)
```
AntonReinhard, line 50: Becomes a little easier to read by leaving the negation out and swapping the results.
```suggestion
    return (isapprox(sum(in_ps), sum(out_ps))) ? _all_onshell(proc, in_ps, out_ps) : false
```

test/processes/one_photon_compton/perturbative.jl
```diff
@@ -0,0 +1,118 @@
...
+                                       Iterators.product(OMEGAS, COS_THETAS, PHIS)
+            IN_COORDS = [om]
+            OUT_COORDS = [cth, phi]
+            IN_PS, OUT_PS = QEDprocesses._generate_momenta(
+                PROC, MODEL, PS_DEF, IN_COORDS, OUT_COORDS
+            )
+            in_mom_square = getMass2.(IN_PS)
+            out_mom_square = getMass2.(OUT_PS)
+            in_masses = mass.(incoming_particles(PROC)) .^ 2
+            out_masses = mass.(outgoing_particles(PROC))
```
AntonReinhard, line 35: Is this supposed to not be squared?

test/processes/one_photon_compton/process.jl
```diff
@@ -0,0 +1,57 @@
+using QEDprocesses
+using Random
+using QEDbase
+using QEDprocesses
```
AntonReinhard, line 4: Already used above
```suggestion
```

test/runtests.jl
```diff
@@ -21,7 +21,11 @@ begin
     @time @safetestset "cross section & probability" begin
         include("cross_sections.jl")
     end
+
     @time @safetestset "phase spaces" begin
         include("phase_spaces.jl")
     end
+
+    # scattering processes
+    include("processes/run_process_test.jl")
```
AntonReinhard, line 30: I think it would still make sense to wrap this in its own `@safetestset`. Then the file structure just represents a depth of the test sets too and this file is more consistent by having each include in a testset.

I'm not sure if the `@time` macros are really useful here, do you use them? The basic time info of how long a testset took to execute is printed anyways, and the allocations/compile times don't really mean anything on their own.

test/processes/one_photon_compton/perturbative.jl
szabo137: You are right, done in bb7f08c.

src/processes/one_photon_compton/one_photon_compton.jl
szabo137: Removed in bb7f08c.

src/patch_QEDbase.jl
szabo137: Yep, that's right, the fix is given here https://github.com/QEDjl-project/QEDbase.jl/pull/63 and a comment is added in bb7f08c.

test/runtests.jl
szabo137: The macro `@safetestset` builds a module around the most outer testset, which makes the tests in there self-contained. I would like to have the tests for each process in a separate model, therefore, I moved the `@safetestset` to the process tests itself. 

Usually, I use the `@time` macro to get more information about the particular testset, e.g. gc time, or recompilation time, not only the full time taken for the test.

szabo137:
As we discussed offline, we want to merge this PR without connecting to the phase space points, and then we want to open a separate PR for the change of the diffCS interface to phase space points (see #57).

@AntonReinhard I added the total cross section, so there are no open todos. I think this is ready for final review.

test/runtests.jl
AntonReinhard: I think that's fair for now. I do hope though that Julia will add some more functionality to its testing in the future. Then we could think about this again but it's not urgent at all.

test/processes/one_photon_compton/perturbative.jl
```diff
@@ -5,19 +5,19 @@ using Random
...
 
 include("groundtruths.jl")
 
-MODEL = PerturbativeQED()
-PS_DEF = PhasespaceDefinition(SphericalCoordinateSystem(), ElectronRestFrame())
-OMEGAS = (1e-6 * rand(RNG), 1e-3 * rand(RNG), rand(RNG), 1e3 * rand(RNG))
+const MODEL = PerturbativeQED()
+const PS_DEF = PhasespaceDefinition(SphericalCoordinateSystem(), ElectronRestFrame())
+const OMEGAS = (1e-6 * rand(RNG), 1e-3 * rand(RNG), rand(RNG), 1e3 * rand(RNG))
 #OMEGAS = (rand(RNG),)
```
AntonReinhard, line 17: This seems leftover?
```suggestion
```

AntonReinhard (approved):

test/processes/one_photon_compton/perturbative.jl
szabo137: done in 87ee685.

szabo137:
Certainly, there was a bug in `_incident_flux` which was only present in Julia versions below 1.10. 

TL;DR: 
In the tested versions (1.6 to 1.9), the function `prod(::Vector{AFM})` consuming a vector of four-vectors returned `Union{Float64,<:StaticArray}`, which is clear since the compiler does not know anything about the number of elements passed to `prod`. However, since for Compton we know that `length(in_ps)==2`, we can just index into the vector and use the mul function.

AntonReinhard (approved):
Looks good

Issue #46, opened 2024-03-27
QEDInput for differential cross-section interface


AntonReinhard:
After some talking with @szabo137 we came up with the following idea for the public differential cross-section interface:

The current interfaces have to check that the input/output phasespaces have the correct dimensionalities to fit to the process definition. We suggest adding a `QEDProcessInput` type that includes all the necessary information of the process in its template parameters and also one point in the phasespace.
Something like
```julia
struct QEDProcessInput{Process, N1, N2, N3, N4, N5, N6}
    proc::Process
    inFerms::SVector{N1, FermionStateful{Incoming}}
    outFerms::SVector{N2, FermionStateful{Outgoing}}
    inAntiferms::SVector{N3, AntiFermionStateful{Incoming}}
    outAntiferms::SVector{N4, AntiFermionStateful{Outgoing}}
    inPhotons::SVector{N5, PhotonStateful{Incoming}}
    outPhotons::SVector{N6, PhotonStateful{Outgoing}}
end
```

Then, a diffCS implementation taking just a `QEDProcessInput` and a model would be provided, assembling the process, in-phase-space-definition, in-phase-space-point, out-phase-space-definition, and out-phase-space-definition from them and dispatching to the internal implementation.
The internal implementations could then be hidden (add _) and would no longer need to provide a safety check for the dimensionalities of the phase spaces.

It would make the usage very simple, by allowing calls such as:
```julia
julia> output = diffCS(model, input)
julia> outputs = diffCS.(model, inputs)
julia> cuOutputs = diffCS.(model, CuArray(inputs))
```

szabo137:
I like the idea! However, there is at least two open questions on my side:

## Type stability
I assume that `FermionStateful` is the combination of the respective fermion momentum, direction, and spin. However, this means, for instance, that the `SVector` in the field `inFerm` contains different element types, if, e.g., the spins are different: assume we have something like this

```julia
using QEDbase
using StaticArrays

struct FermionStateful{D<:ParticleDirection, S<:AbstractSpin}
  mom::SFourMomentum
end
```

Then we quickly run into performance issues:

setup

```julia

mom = rand(SFourMomentum)
s1 = FermionStateful{Imcoming,SpinUp}(mom)
s2 = FermionStateful{Imcoming,SpinDown}(mom)

v_stable = SVector(s1,s1)
v_unstable = SVector(s1,s2)
```

benchmark: indexing

```julia
using BenchmarkTools

[...]

PR #51, opened 2024-05-03
Add PhaseSpacePoint definition

AntonReinhard:
Proposal to fix issue #46

This branch is currently based on PR #39 and should be merged after it.

This is a draft for now, but I think it is a good compromise from the discussion in issue #46. The definition is type stable, allows some nice overloads for `is_fermion` and so on for the `ParticleStateful` objects, and takes care of preventing illegal objects in the constructors.

Creating particles is straightforward and disallows illegal configurations (electron with polarization or photon with spin):
```julia
in_el = ParticleStateful(rand(SFourMomentum), Electron(), AllSpin(), Incoming())
in_ph = ParticleStateful(rand(SFourMomentum), Photon(), AllPol(), Incoming())
out_el = ParticleStateful(rand(SFourMomentum), Electron(), AllSpin(), Outgoing())
out_ph = ParticleStateful(rand(SFourMomentum), Photon(), AllPol(), Outgoing())

in_particles_valid = SVector(in_el, in_ph)
in_particles_invalid = SVector(in_el, out_ph)

out_particles_valid = SVector(out_el, out_ph)
out_particles_invalid = SVector(out_el, in_ph)
```

Creating `PhaseSpacePoint`s is similarly simple and prevents illegal configurations, e.g. incoming particles in the outgoing list, but is still reasonably fast and does not need allocations. One PhaseSpacePoint for a 2 to 2 process is 944B in memory.
```julia
julia> @benchmark PhaseSpacePoint($process, $model, $phasespace_def, $in_particles_valid, $out_particles_valid)
BenchmarkTools.Trial: 10000 samples with 991 evaluations.
 Range (min … max):  39.268 ns … 111.740 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     41.068 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   43.116 ns ±   6.557 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▃▇█▆▂ ▁▂▁▁                                             ▁    ▁
  ▆███████████▇▆▆▇▆▆▆▅▅▆▆▆▅▄▅▅▄▄▆▆▆▆▆▆▇▇▆▇▇█▇▅▄▅▅▄▅▄▅▄▅▆████▇▇ █
  39.3 ns       Histogram: log(frequency) by time      70.9 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

Suggestions welcome

### DIFF
--- a/src/phase_spaces.jl
+++ b/src/phase_spaces.jl
@@ -224,3 +224,11 @@ end
 @inline is_anti_particle(particle::ParticleStateful) = is_anti_particle(particle.species)
 @inline mass(particle::ParticleStateful) = mass(particle.species)
 @inline charge(particle::ParticleStateful) = charge(particle.species)
+
+@inline _spin(::Species, particle::ParticleStateful) where {Species<:FermionLike} =
+    particle.spin_or_pol
+@inline spin(particle::ParticleStateful) = _spin(particle.species, particle)
+
+@inline _pol(::Species, particle::ParticleStateful) where {Species<:BosonLike} =
+    particle.spin_or_pol
+@inline pol(particle::ParticleStateful) = _pol(particle.species, particle)
--- a/src/phase_spaces.jl
+++ b/src/phase_spaces.jl
@@ -15,6 +15,21 @@ import QEDbase.is_fermion
 import QEDbase.is_boson
 import QEDbase.is_incoming
 import QEDbase.is_outgoing
+import QEDbase.mass
+import QEDbase.charge
+
+import QEDbase:
+    is_particle,
+    is_anti_particle,
+    is_fermion,
+    is_boson,
+    is_incoming,
+    is_outgoing,
+    mass,
+    charge,
+    AbstractFourMomentum,
+    AbstractParticleType,
+    AbstractParticle
 
 abstract type AbstractCoordinateSystem end
 struct SphericalCoordinateSystem <: AbstractCoordinateSystem end
@@ -41,7 +56,7 @@ end
 #
 # Currently, elements can be either four-momenta, or real numbers,
 # i.e. coordinates.
-AbstractPhasespaceElement = Union{QEDbase.AbstractFourMomentum,Real}
+AbstractPhasespaceElement = Union{AbstractFourMomentum,Real}
 
 # utility functions 
 
@@ -49,7 +64,7 @@ AbstractPhasespaceElement = Union{QEDbase.AbstractFourMomentum,Real}
     proc::AbstractProcessDefinition,
     ::AbstractModelDefinition,
     in_phase_space::AbstractVecOrMat{T},
-) where {T<:QEDbase.AbstractFourMomentum}
+) where {T<:AbstractFourMomentum}
     return size(in_phase_space, 1) == number_incoming_particles(proc) || throw(
         DimensionMismatch(
             "The number of incoming particles <$(number_incoming_particles(proc))> is inconsistent with input size <$(size(in_phase_space,1))>",
@@ -61,7 +76,7 @@ end
     proc::AbstractProcessDefinition,
     ::AbstractModelDefinition,
     out_phase_space::AbstractVecOrMat{T},
-) where {T<:QEDbase.AbstractFourMomentum}
+) where {T<:AbstractFourMomentum}
     return size(out_phase_space, 1) == number_outgoing_particles(proc) || throw(
         DimensionMismatch(
             "The number of outgoing particles <$(number_outgoing_particles(proc))> is inconsistent with input size <$(size(out_phase_space,1))>",
@@ -94,45 +109,36 @@ end
 end
 
 """
-    ParticleStateful
+    ParticleStateful <: AbstractParticle
 
 Representation of a particle with a state. It has four fields:
-- `mom::AbstractPhasespaceElement`: The momentum of the particle.
-- `type::AbstractParticle`: The type of the particle, `QEDbase.Electron()`, `QEDbase.Positron()` etc..
-- `spinorpol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc.. Can only use spins with fermions and polarizations with bosons.
 - `dir::ParticleDirection`: The direction of the particle, `QEDbase.Incoming()` or `QEDbase.Outgoing()`.
+- `species::AbstractParticleType`: The species of the particle, `QEDbase.Electron()`, `QEDbase.Positron()` etc..
+- `mom::AbstractFourMomentum`: The momentum of the particle.
+- `spin_or_pol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc.. Can only use spins with fermions and polarizations with bosons.
 
-Overloads for `QEDbase.is_fermion`, `QEDbase.is_boson`, `QEDbase.is_particle`, `QEDbase.is_anti_particle`, `QEDbase.is_incoming`, and `QEDbase.is_outgoing` are provided, delegating the call to the correct field.
+Overloads for `QEDbase.is_fermion`, `QEDbase.is_boson`, `QEDbase.is_particle`, `QEDbase.is_anti_particle`, `QEDbase.is_incoming`, `QEDbase.is_outgoing`, `QEDbase.mass`, and `QEDbase.charge` are provided, delegating the call to the correct field and thus implementing the `QEDbase.AbstractParticle` interface.
 
 The legality of the combination of `type` and `spinorpol` is checked on construction. If, for example, the construction of an `Electron()` with a polarization is attempted, an [`InvalidInputError`](@ref) is thrown.
 """
-struct ParticleStateful{ElType<:AbstractPhasespaceElement}
-    mom::ElType
-    type::QEDbase.AbstractParticle
-    spinorpol::AbstractSpinOrPolarization
+struct ParticleStateful{ElType<:AbstractFourMomentum} <: AbstractParticle
     dir::ParticleDirection
+    species::AbstractParticleType
+    mom::ElType
+    spin_or_pol::AbstractSpinOrPolarization
 
     function ParticleStateful(
-        mom::ElType,
-        type::QEDbase.AbstractParticle,
-        spinorpol::AbstractSpinOrPolarization,
-        dir::ParticleDirection,
-    ) where {ElType<:AbstractPhasespaceElement}
-        if is_fermion(type) && !(spinorpol isa AbstractSpin)
-            throw(
-                InvalidInputError(
-                    "Tried to construct a stateful fermion with a $(spinorpol) instead of a spin",
-                ),
-            )
-        elseif is_boson(type) && !(spinorpol isa AbstractPolarization)
-            throw(
-                InvalidInputError(
-                    "Tried to construct a stateful boson with a $(spinorpol) instead of a polarization",
-                ),
-            )
-        end
+        dir::ParticleDirection, species::Species, mom::ElType, spin::Spin=AllSpin()
+    ) where {Species<:FermionLike,ElType<:AbstractFourMomentum,Spin<:AbstractSpin}
+        # constructor for fermions with spin
+        return new{ElType}(dir, species, mom, spin)
+    end
 
-        return new{ElType}(mom, type, spinorpol, dir)
+    function ParticleStateful(
+        dir::ParticleDirection, species::Species, mom::ElType, pol::Pol=AllPol()
+    ) where {Species<:BosonLike,ElType<:AbstractFourMomentum,Pol<:AbstractPolarization}
+        # constructor for bosons with polarization
+        return new{ElType}(dir, species, mom, pol)
     end
 end
 
@@ -147,7 +153,7 @@ struct PhaseSpacePoint{
     PROC<:AbstractProcessDefinition,
     MODEL<:AbstractModelDefinition,
     PSDEF<:AbstractPhasespaceDefinition,
-    PhaseSpaceElementType<:AbstractPhasespaceElement,
+    PhaseSpaceElementType<:AbstractFourMomentum,
     N_IN_PARTICLES,
     N_OUT_PARTICLES,
 }
@@ -164,7 +170,7 @@ struct PhaseSpacePoint{
         PROC<:AbstractProcessDefinition,
         MODEL<:AbstractModelDefinition,
         PSDEF<:AbstractPhasespaceDefinition,
-        PhaseSpaceElementType<:AbstractPhasespaceElement,
+        PhaseSpaceElementType<:AbstractFourMomentum,
         IN_P<:AbstractVector{ParticleStateful{PhaseSpaceElementType}},
         OUT_P<:AbstractVector{ParticleStateful{PhaseSpaceElementType}},
     }
@@ -180,9 +186,9 @@ struct PhaseSpacePoint{
         )
 
         for (proc_p, p) in zip(incoming_particles(proc), in_p)
-            proc_p == p.type || throw(
+            proc_p == p.species || throw(
                 InvalidInputError(
-                    "Process given particle type ($(proc_p)) does not match stateful particle type ($(p.type))",
+                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species))",
                 ),
             )
             is_incoming(p) || throw(
@@ -192,9 +198,9 @@ struct PhaseSpacePoint{
             )
         end
         for (proc_p, p) in zip(outgoing_particles(proc), out_p)
-            proc_p == p.type || throw(
+            proc_p == p.species || throw(
                 InvalidInputError(
-                    "Process given particle type ($(proc_p)) does not match stateful particle type ($(p.type))",
+                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species))",
                 ),
             )
             is_outgoing(p) || throw(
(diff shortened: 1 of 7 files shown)


test/phase_spaces.jl
```diff
@@ -0,0 +1,76 @@
...
+
+            @test particle_stateful.mom == momentum
+            @test is_fermion(particle_stateful) == is_fermion(particle)
+            @test is_boson(particle_stateful) == is_boson(particle)
+            @test is_particle(particle_stateful) == is_particle(particle)
+            @test is_anti_particle(particle_stateful) == is_anti_particle(particle)
+            @test is_incoming(particle_stateful) == is_incoming(dir)
+            @test is_outgoing(particle_stateful) == is_outgoing(dir)
+        else
+            @test_throws "Tried to construct a stateful $(is_fermion(particle) ? "fermion" : "boson") with a $(spinorpol) instead of a $(is_fermion(particle) ? "spin" : "polarization")" ParticleStateful(
```
szabo137, line 33: This checks the throwing of a string, but an should check for an `InvalidInputError`.

test/phase_spaces.jl
```diff
@@ -0,0 +1,76 @@
...
+    model = TESTMODEL
+    process = TestImplementation.TestProcess(
+        SVector{2,AbstractParticle}(Electron(), Photon()),
+        SVector{2,AbstractParticle}(Electron(), Photon()),
+    )
+    phasespace_def = TESTPSDEF
+
+    PhaseSpacePoint(process, model, phasespace_def, in_particles_valid, out_particles_valid)
+
+    @test_throws r"Stateful particle (.*) is given as an incoming particle but is outgoing" PhaseSpacePoint(
```
szabo137, line 61: Same as above. This should check of a `InvalidInputError` is thrown. 

Same for the next three tests.

test/phase_spaces.jl
```diff
@@ -0,0 +1,76 @@
...
+TESTMODEL = TestImplementation.TestModel()
+TESTPSDEF = TestImplementation.TestPhasespaceDef()
+
+@testset "Stateful Particle" begin
+    DIRECTIONS = [Incoming(), Outgoing()]
+    PARTICLES = [Electron(), Positron()] #=, Muon(), AntiMuon(), Tauon(), AntiTauon()=#
+    SPINANDPOLS = [AllSpin(), SpinUp(), SpinDown(), AllPol(), PolX(), PolY()]
+
+    for (particle, dir, spinorpol) in Iterators.product(PARTICLES, DIRECTIONS, SPINANDPOLS)
+        momentum = rand(SFourMomentum)
```
szabo137, line 19: For reproducibility, you should use an seeded RNG here, e.g. `Random.MersenneTwister(123456789)`.

szabo137 (commented):
Thank you for this! 

This is not a full review of the changes, but some first obvious comments from my side. 

Can you explain, why the unit tests for Julia v1.6 and v1.7 fail, but the others not? Looks a bit like an API change for `@test_throws`.

AntonReinhard:
> Can you explain, why the unit tests for Julia v1.6 and v1.7 fail, but the others not? Looks a bit like an API change for `@test_throws`.

Yeah I'm pretty sure that's what it is... I think the matching to a regex is new. I can just wrap it in a check about the Julia version, but are we sure we need to support 1.6 and 1.7 anyways? Especially when the now stable version is 1.10 already.

test/phase_spaces.jl
AntonReinhard: I don't think it's possible to do both at once, other than with a matching function. I just duplicated the tests now and do the regex matches only when the Julia version is 1.8 or newer.

szabo137 (commented):
I like the implementation, here are some comments from my side. 

Furthermore, we should think about some custom `Base.show` methods for these nested types.

szabo137:
> > Can you explain, why the unit tests for Julia v1.6 and v1.7 fail, but the others not? Looks a bit like an API change for `@test_throws`.
> 
> Yeah I'm pretty sure that's what it is... I think the matching to a regex is new. I can just wrap it in a check about the Julia version, but are we sure we need to support 1.6 and 1.7 anyways? Especially when the now stable version is 1.10 already.

Julia v1.6 is the defacto LTS, but you are right, the tests for v1.7 to v1.9 are somewhat arbitrary. However, I am fine with the exclusion of some tests, if there is an API change of lack of support in the respective Julia version. Therefore, there is no need to drop the support in the current status, but we should keep an eye on that. 

Regarding the support of Julia v1.10, you are right, of course. There should be an issue opened on its inclusion 😉

src/phase_spaces.jl
```diff
@@ -83,3 +92,127 @@ end
...
+"""
+struct ParticleStateful{ElType<:AbstractPhasespaceElement}
+    mom::ElType
+    type::QEDbase.AbstractParticle
+    spinorpol::AbstractSpinOrPolarization
+    dir::ParticleDirection
+
+    function ParticleStateful(
+        mom::ElType,
+        type::QEDbase.AbstractParticle,
```
szabo137, line 117: `QEDbase.AbstractParticleType`?

src/phase_spaces.jl
```diff
@@ -83,3 +92,127 @@ end
...
+- `spinorpol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc.. Can only use spins with fermions and polarizations with bosons.
+- `dir::ParticleDirection`: The direction of the particle, `QEDbase.Incoming()` or `QEDbase.Outgoing()`.
+
+Overloads for `QEDbase.is_fermion`, `QEDbase.is_boson`, `QEDbase.is_particle`, `QEDbase.is_anti_particle`, `QEDbase.is_incoming`, and `QEDbase.is_outgoing` are provided, delegating the call to the correct field.
+
+The legality of the combination of `type` and `spinorpol` is checked on construction. If, for example, the construction of an `Electron()` with a polarization is attempted, an [`InvalidInputError`](@ref) is thrown.
+"""
+struct ParticleStateful{ElType<:AbstractPhasespaceElement}
+    mom::ElType
+    type::QEDbase.AbstractParticle
```
szabo137, line 111: `QEDbase.AbstractParticleType`?

src/phase_spaces.jl
```diff
@@ -83,3 +92,127 @@ end
...
+- `mom::AbstractPhasespaceElement`: The momentum of the particle.
+- `type::AbstractParticle`: The type of the particle, `QEDbase.Electron()`, `QEDbase.Positron()` etc..
+- `spinorpol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc.. Can only use spins with fermions and polarizations with bosons.
+- `dir::ParticleDirection`: The direction of the particle, `QEDbase.Incoming()` or `QEDbase.Outgoing()`.
+
+Overloads for `QEDbase.is_fermion`, `QEDbase.is_boson`, `QEDbase.is_particle`, `QEDbase.is_anti_particle`, `QEDbase.is_incoming`, and `QEDbase.is_outgoing` are provided, delegating the call to the correct field.
+
+The legality of the combination of `type` and `spinorpol` is checked on construction. If, for example, the construction of an `Electron()` with a polarization is attempted, an [`InvalidInputError`](@ref) is thrown.
+"""
+struct ParticleStateful{ElType<:AbstractPhasespaceElement}
```
szabo137, line 109: Maybe we should subtype `<:QEDbase.AbstractParticle` and implement the rest of the interface, e.g. `mass` and `charge`.

src/phase_spaces.jl
```diff
@@ -83,3 +92,127 @@ end
         ),
     )
 end
+
+"""
+    ParticleStateful
```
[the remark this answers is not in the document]
AntonReinhard: As discussed in person, I believe it's a little more accurate to use 'stateful' instead of just 'state', since it is the particle and the state, not only the state. Also, I prefer `ParticleStateful` over `StatefulParticle` because I think it will be more useful for autocomplete, as I'd start typing Particle before thinking of the state.

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+    Base.getindex(psp::PhaseSpacePoint, dir::Outgoing, n::Int)
+
+Overload for the array indexing operator `[]`. Returns the nth outgoing particle in this phase space point.
+"""
+function Base.getindex(psp::PhaseSpacePoint, ::Outgoing, n::Int)
+    return psp.out_particles[n]
+end
+
+"""
+    nth_momentum(psp::PhaseSpacePoint, dir::ParticleDirection, n::Int)
```
szabo137, line 255: The naming is not very idiomatic: I suggest using just `momentum(psp, dir, n)`.

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+@inline mass(particle::ParticleStateful) = mass(particle.species)
+@inline charge(particle::ParticleStateful) = charge(particle.species)
+
+@inline _spin(::Species, particle::ParticleStateful) where {Species<:FermionLike} =
+    particle.spin_or_pol
+@inline spin(particle::ParticleStateful) = _spin(particle.species, particle)
+
+@inline _pol(::Species, particle::ParticleStateful) where {Species<:BosonLike} =
+    particle.spin_or_pol
+@inline pol(particle::ParticleStateful) = _pol(particle.species, particle)
```
szabo137, line 160: For internal use, abbreviations like `pol` are fine, but I suggest using `polarization(particle_stateful)` for an exported function.

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+
+"""
+    PhaseSpacePoint
+
+Representation of a point in the phase space of a process. Contains the process ([`AbstractProcessDefinition`](@ref)), the model ([`AbstractModelDefinition`](@ref)), the phase space definition ([`AbstractPhasespaceDefinition`]), and stateful incoming and outgoing particles ([`ParticleStateful`](@ref)).
+
+The legality of the combination of the given process and the incoming and outgoing particles is checked on construction. If the numbers of particles mismatch, the types of particles mismatch (note that order is important), or incoming particles have a `Outgoing` direction, an error is thrown.
```
szabo137, line 167:
```suggestion
The legality of the combination of the given process and the incoming and outgoing particles is checked on construction. If the numbers of particles mismatch, the types of particles mismatch (note that order is important), or incoming particles have an `Outgoing` direction, an error is thrown.
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+        PROC<:AbstractProcessDefinition,
+        MODEL<:AbstractModelDefinition,
+        PSDEF<:AbstractPhasespaceDefinition,
+        PhaseSpaceElementType<:AbstractFourMomentum,
+        IN_P<:AbstractVector{ParticleStateful{PhaseSpaceElementType}},
+        OUT_P<:AbstractVector{ParticleStateful{PhaseSpaceElementType}},
+    }
+        length(incoming_particles(proc)) == length(in_p) || throw(
+            InvalidInputError(
+                "The number of incoming particles given by the process ($(incoming_particles(proc))) mismatches the number of given stateful incoming particles ($(length(in_p)))",
```
szabo137, line 196:
```suggestion
                "The number of incoming particles given by the process ($(incoming_particles(proc))) mismatches the number of given stateful incoming particles ($(length(in_p))).",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+        length(outgoing_particles(proc)) == length(out_p) || throw(
+            InvalidInputError(
+                "The number of outgoing particles given by the process ($(outgoing_particles(proc))) mismatches the number of given stateful outgoing particles ($(length(out_p)))",
+            ),
+        )
+
+        for (proc_p, p) in zip(incoming_particles(proc), in_p)
+            proc_p == p.species || throw(
+                InvalidInputError(
+                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species))",
```
szabo137, line 208:
```suggestion
                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species)).",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+                ),
+            )
+            is_incoming(p) || throw(
+                InvalidInputError(
+                    "Stateful particle $(p) is given as an incoming particle but is outgoing",
```
szabo137, line 213:
```suggestion
                    "Stateful particle $(p) is given as an incoming particle but is outgoing.",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+                ),
+            )
+        end
+        for (proc_p, p) in zip(outgoing_particles(proc), out_p)
+            proc_p == p.species || throw(
+                InvalidInputError(
+                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species))",
```
szabo137, line 220:
```suggestion
                    "Process given particle species ($(proc_p)) does not match stateful particle species ($(p.species)).",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+                ),
+            )
+            is_outgoing(p) || throw(
+                InvalidInputError(
+                    "Stateful particle $(p) is given as an outgoing particle but is incoming",
```
szabo137, line 225:
```suggestion
                    "Stateful particle $(p) is given as an outgoing particle but is incoming.",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+Overload for the array indexing operator `[]`. Returns the nth outgoing particle in this phase space point.
+"""
+function Base.getindex(psp::PhaseSpacePoint, ::Outgoing, n::Int)
+    return psp.out_particles[n]
+end
+
+"""
+    nth_momentum(psp::PhaseSpacePoint, dir::ParticleDirection, n::Int)
+
+Returns the momentum of the `n`th particle in the given [`PhaseSpacePoint`](@ref) which has direction `dir`. If `n` is outside the valid range for this phase space point, an `BoundsError` is thrown.
```
szabo137, line 257:
```suggestion
Returns the momentum of the `n`th particle in the given [`PhaseSpacePoint`](@ref) which has direction `dir`. If `n` is outside the valid range for this phase space point, a `BoundsError` is thrown.
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+        OUT_P<:AbstractVector{ParticleStateful{PhaseSpaceElementType}},
+    }
+        length(incoming_particles(proc)) == length(in_p) || throw(
+            InvalidInputError(
+                "The number of incoming particles given by the process ($(incoming_particles(proc))) mismatches the number of given stateful incoming particles ($(length(in_p)))",
+            ),
+        )
+        length(outgoing_particles(proc)) == length(out_p) || throw(
+            InvalidInputError(
+                "The number of outgoing particles given by the process ($(outgoing_particles(proc))) mismatches the number of given stateful outgoing particles ($(length(out_p)))",
```
szabo137, line 201:
```suggestion
                "The number of outgoing particles given by the process ($(outgoing_particles(proc))) mismatches the number of given stateful outgoing particles ($(length(out_p))).",
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
         ),
     )
 end
+
+"""
+    ParticleStateful <: AbstractParticle
+
+Representation of a particle with a state. It has four fields:
+- `dir::ParticleDirection`: The direction of the particle, `QEDbase.Incoming()` or `QEDbase.Outgoing()`.
+- `species::AbstractParticleType`: The species of the particle, `QEDbase.Electron()`, `QEDbase.Positron()` etc..
```
szabo137, line 116:
```suggestion
- `species::AbstractParticleType`: The species of the particle, `QEDbase.Electron()`, `QEDbase.Positron()` etc.
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+- `mom::AbstractFourMomentum`: The momentum of the particle.
+- `spin_or_pol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc.. Can only use spins with fermions and polarizations with bosons.
```
szabo137, line 118:
```suggestion
- `spin_or_pol::AbstractSpinOrPolarization`: The spin or polarization of the particle, `QEDbase.SpinUp()`, `QEDbase.PolX() etc. Can only use spins with fermions and polarizations with bosons.
```

src/phase_spaces.jl
```diff
@@ -83,3 +107,155 @@ end
...
+
+Overloads for `QEDbase.is_fermion`, `QEDbase.is_boson`, `QEDbase.is_particle`, `QEDbase.is_anti_particle`, `QEDbase.is_incoming`, `QEDbase.is_outgoing`, `QEDbase.mass`, and `QEDbase.charge` are provided, delegating the call to the correct field and thus implementing the `QEDbase.AbstractParticle` interface.
+
+The legality of the combination of `type` and `spinorpol` is checked on construction. If, for example, the construction of an `Electron()` with a polarization is attempted, an [`InvalidInputError`](@ref) is thrown.
```
szabo137, line 122:
```suggestion
The legality of the combination of `species` and `spin_or_pol` is checked on construction. If, for example, the construction of an `Electron()` with a polarization is attempted, an [`InvalidInputError`](@ref) is thrown.
```

szabo137 (approved):
Just some minor issues. Can be merged if resolved.

src/phase_spaces.jl
AntonReinhard: I just decided to look up what error messages are supposed to look like according to Julia guidelines and found this from [here](https://docs.julialang.org/en/v1/manual/control-flow/#Exception-Handling)
```
When writing an error message, it is preferred to make the first word lowercase. For example,

size(A) == size(B) || throw(DimensionMismatch("size of A not equal to size of B"))

is preferred over

size(A) == size(B) || throw(DimensionMismatch("Size of A not equal to size of B")).

However, sometimes it makes sense to keep the uppercase first letter, for instance if an argument to a function is a capital letter:

size(A,1) == size(B,2) || throw(DimensionMismatch("A has first dimension...")).
```

So I think we should actually follow that. The `InvalidInputError`'s `showerror` function also adds its own period at the end of the message, which is why I left it out before, but it probably should not do that either.

szabo137: You are right. My bad 🤷‍♂️

szabo137:
Btw: we must not set assignees on pull requests, because it breaks the gitlab bot somehow 🤷‍♂️

AntonReinhard:
Oh you're right, I completely forgot

PR #52, opened 2024-05-06
cross sections on phase space points

szabo137:
With this PR, we adopt the phase space points from #51 for the probability and cross section interface.

Further changes: we drop the support of different phase space definitions for the in- and out-phase space. 

⚠️ This is based on https://github.com/AntonReinhard/QEDprocesses.jl/tree/phase_space_point and should be rebased to `dev` and considered for merging *after* #51 is merged.


src/phase_spaces.jl
```diff
@@ -159,6 +161,17 @@ end
...
+const PHASESPACE_VALIDITY_CHECK= Ref(true)
+
+macro valid_phase_space(ex)
+    return quote
+        PHASESPACE_VALIDITY_CHECK.x = false
+        local val = $(esc(ex))
+        PHASESPACE_VALIDITY_CHECK.x = true
+        val
+    end
+end
```
AntonReinhard, line 173: What part of this will be part of the public interface? I'm assuming `@valid_phase_space`, in which case it should definitely get some documentation, and maybe the constant too with some sort of warning to not mess with it and/or an _ at the beginning.

Also, I'm not sure if this is the best way of disabling the check. I get the idea, but assuming more complex applications of this in the future, it could happen that there are nested calls to this macro, in which case unexpected things will happen. That could of course be fixed by checking if the check is already turned off before turning it off as a recursion check, but I'm wondering if there isn't a better way. I don't really have a better suggestion myself right now though.

src/phase_spaces.jl
```diff
@@ -226,6 +240,7 @@ struct PhaseSpacePoint{
                 ),
             )
         end
+    end
```
AntonReinhard, line 243: Please fix the indentation

src/probabilities.jl
```diff
@@ -87,32 +82,39 @@ end
...
     return res
 end
 
+function unsafe_differential_probability(phase_space_point::PhaseSpacePoint)
+  return _unsafe_differential_probability(
+    phase_space_point.proc,
+    phase_space_point.model,
+    phase_space_point.ps_def,
+    momentum.(phase_space_point.in_particles),
+    momentum.(phase_space_point.out_particles)
```
AntonReinhard, line 108: I think you're assuming a function like
```julia
momentum(p::ParticleStateful) = p.mom
```
but that doesn't exist currently?

AntonReinhard: Interestingly, assuming the above function exists, `momentum.(psp.in_particles)` is orders of magnitude faster than `getfield.(psp.in_particles, :mom)` according to `@benchmark`. It seems it's a bad idea to directly broadcast with `getfield()`

src/phase_spaces.jl
szabo137: I agree with your reasoning. Therefore, I removed this.

src/probabilities.jl
szabo137: Thats right! I guess something is under the hood, because `getfield` is from `Core`, i.e. not written in Julia itself, and therefore not suitable for broadcasting.

szabo137:
@AntonReinhard I think this is ready for review. 

Currently, there is a QEDbase.jl patch in here. I think we should include this in QEDbase.jl, but the release would take a while. What do you think, is it valid to have `patch_something.jl` files in there, and remove the functionality, after the respective functions are released?

AntonReinhard:
> @AntonReinhard I think this is ready for review.
> 
> Currently, there is a QEDbase.jl patch in here. I think we should include this in QEDbase.jl, but the release would take a while. What do you think, is it valid to have `patch_something.jl` files in there, and remove the functionality, after the respective functions are released?
> 
> Edit: The respective PR is [QEDjl-project/QEDbase.jl#62](https://github.com/QEDjl-project/QEDbase.jl/pull/62)

I think this is a good idea, since the alternative is either wait until the patch is merged into the dependency and released, or depending on a non-released state of the dependency. I think the former is slow and the latter is ugly.
We could maybe make it a rule though to have a reference to an opened issue/PR in the first lines of each `patch_something.jl` file. That should prevent the issues getting lost.

szabo137:
> > @AntonReinhard I think this is ready for review.
> > Currently, there is a QEDbase.jl patch in here. I think we should include this in QEDbase.jl, but the release would take a while. What do you think, is it valid to have `patch_something.jl` files in there, and remove the functionality, after the respective functions are released?
> > Edit: The respective PR is [QEDjl-project/QEDbase.jl#62](https://github.com/QEDjl-project/QEDbase.jl/pull/62)
> 
> I think this is a good idea, since the alternative is either wait until the patch is merged into the dependency and released, or depending on a non-released state of the dependency. I think the former is slow and the latter is ugly. We could maybe make it a rule though to have a reference to an opened issue/PR in the first lines of each `patch_something.jl` file. That should prevent the issues getting lost.

I think this is a great idea! I added the respective information in 86beefe.

AntonReinhard (approved):
Looks good to me

A general remark though: Some of the test files (cross_sections.jl for example) are getting pretty long and deeply nested. It might be a good idea to move some of the nesting into a directory/file structure, because it's becoming hard to keep track of what level a testset should actually be at. Should probably make that a separate issue though.

AntonReinhard:
The first two commits in this branch are from me and have no total changes, you might wanna remove them from the branch history. But I guess we squash anyways so it's not important.

szabo137:
> The first two commits in this branch are from me and have no total changes, you might wanna remove them from the branch history. But I guess we squash anyways so it's not important.

Right, since the commits are squashed anyways, this is not an issue.

szabo137:
> Looks good to me
> 
> 
> 
> A general remark though: Some of the test files (cross_sections.jl for example) are getting pretty long and deeply nested. It might be a good idea to move some of the nesting into a directory/file structure, because it's becoming hard to keep track of what level a testset should actually be at. Should probably make that a separate issue though.

I like the idea and opened an issue: https://github.com/QEDjl-project/QEDprocesses.jl/issues/56

PR #55, opened 2024-05-08
Proposal: Benchmarks for phase space points

szabo137:
TBW


benchmark/plot_bench.jl
```diff
@@ -0,0 +1,36 @@
...
+using BenchmarkPlots, StatsPlots
+
+plotpath = "plots"
+if !isdir(plotpath)
+    mkdir(plotpath)
+end
+
+result = BenchmarkTools.load("bench.json")[1]
+
+P = plot(result["stateful"]["accessor"]; yscale=:log10, ylim=(1e-1, 100))
```
AntonReinhard, line 14: Is there a reason for not using the automatic y-limits? Executing this on my machine (and probably many others) will make the values go out of these ranges.

AntonReinhard (commented):
I like the idea!

benchmark/plot_bench.jl
szabo137: You are right, this is not practical at all. In f0847c34685a8d274cab7f41204d7ef1c53e7341 I added a simple ylim finder to adjust the lims automatically. 

I'm not particularly eager to use the default ylims, because they tend to produce non-integer powers of 10 in log scale.

AntonReinhard: > I'm not particularly eager to use the default ylims, because they tend to produce non-integer powers of 10 in log scale.

Ah right that makes sense.

benchmark/test_implementation.jl
```diff
@@ -0,0 +1,34 @@
...
+end
+
+function TestProcess(rng::AbstractRNG, N_in::Int, N_out::Int)
+    in_particles = rand(rng, PARTICLE_SET, N_in)
+    out_particles = rand(rng, PARTICLE_SET, N_out)
+    return TestProcess(in_particles, out_particles)
+end
+
+QEDprocesses.incoming_particles(proc::TestProcess) = proc.incoming_particles
+QEDprocesses.outgoing_particles(proc::TestProcess) = proc.outgoing_particles
```
AntonReinhard, line 29: According to the interface description, these should return tuples, but currently they return Vectors, which is bad because it's type unstable.

AntonReinhard: The same problem exists in the TestImplementation of the tests themselves. I changed this in #63

PR #60, opened 2024-05-16
[WIP] Pretty printing

AntonReinhard:
Adds pretty printing for:
- PerturbativeQED model
- Compton process
- Coordinate systems, frames of reference, and phase space definition
- Particle stateful
- PhaseSpacePoint

Essentially, adding similar functionality to https://github.com/QEDjl-project/QEDbase.jl/pull/61 for the QEDprocesses types.

I'm using the jldoctests to test, I'm not sure if it's really necessary to test in a separate test as well. Putting single jldoctests for these prints in every type definition would clutter the documentation a bit so I've not done that. If we really want tests for each `show()` we should probably do it in a test file.

Also, I'm not sure if there's a better way for the print functions than alternating the `show` for objects and `print` for strings. Interpolating doesn't work unless we also overload the `print`. (I *think* it is print, the number of string output related functions in Julia is confusing)


docs/Project.toml
```diff
@@ -1,3 +1,4 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+QEDbase = "10e22c08-3ccb-4172-bfcf-7d7aa3d04d93"
```
szabo137, line 3: Why do we need that here?

src/phase_spaces.jl
```diff
@@ -43,17 +43,46 @@ abstract type AbstractPhasespaceDefinition end
 Broadcast.broadcastable(ps_def::AbstractPhasespaceDefinition) = Ref(ps_def)
 
 """
-
     PhasespaceDefinition(coord_sys::AbstractCoordinateSystem, frame::AbstractFrameOfReference)
 
 Convenient type to dispatch on coordiante systems and frames of reference.
+
+```jldoctest
```
szabo137, line 50: IMHO, we should use `jldoctest` to verify the correctness of examples in the docstring of a function or type, not for unit testing. As you already mentioned: covering all cases in a doc-test causes some clunkiness in the docs, which is why we should have dedicated unit tests, even if we have some duplications.

src/phase_spaces.jl
```diff
@@ -43,17 +43,46 @@ abstract type AbstractPhasespaceDefinition end
...
+end
+function Base.show(io::IO, ::MIME"text/plain", ::CenterOfMomentumFrame)
+    print(io, "center-of-momentum frame")
+    return nothing
+end
+function Base.show(io::IO, ::MIME"text/plain", ::ElectronRestFrame)
+    print(io, "electron rest frame")
+    return nothing
+end
+function Base.show(io::IO, m::MIME"text/plain", ps_def::PhasespaceDefinition)
```
szabo137, line 78: I understand the idea behind the more fine granular control over `show` for composite types. However, I don't see the benefit of printing such a type as a one-line English sentence. Regarding the Julia docs, the three-argument version of `show` should display a more formatted type description, whereas the one-liner should be for `print`. In this case here, I thought more about this: 

```julia
julia> PhasespaceDefinition(SphericalCoordinates(),ElectronRestFrame())
PhasespaceDefinition
    coordinate system: spherical coordinates
    frame: electron rest frame
```

Furthermore, I suggest using a different implementation for an inline print (e.g. the two-argument `show`):

```Julia
julia> ps_def = PhasespaceDefinition(SphericalCoordinates(),ElectronRestFrame());
julia> println(ps_def)
spherical coordinates in electron rest frame
```

src/phase_spaces.jl
```diff
@@ -167,12 +206,46 @@ momentum(part::ParticleStateful) = part.mom
     particle.spin_or_pol
 @inline polarization(particle::ParticleStateful) = _polarization(particle.species, particle)
 
+function Base.show(io::IO, m::MIME"text/plain", particle::ParticleStateful)
```
szabo137, line 209: The same as above, I think a more formatted output would be helpful, e.g.

```julia
julia> particle_stateful = ParticleStateful(Outgoing(), Photon(), SFourMomentum(1, 0, 0, 0), PolX());
julia> particle_stateful
ParticleStateful: outgoing photon
    pol: x-polarization
    momentum: [1, 0, 0, 0]
```
The respective inline-print could look like

```julia
julia> println(particle_stateful)
outgoing photon (all pols) [1, 0, 0, 0]
```

Maybe one could also hide the momentum and pol information in the inline print, such that the usage is even more concise.

src/processes/one_photon_compton/process.jl
```diff
@@ -59,3 +59,16 @@ end
 function _spin_or_pol(process::Compton, ::Photon, ::Outgoing)
     return process.out_pol
 end
+
+function Base.show(io::IO, m::MIME"text/plain", proc::Compton)
```
szabo137, line 63: Here, one could think about having two different implementations for formatted block and inline prints, where inline, most of the information could be hidden. 

Example: 

```Julia 
julia> proc = Compton();
julia> proc
one-photon Compton scattering
    incoming: electron (all spin), photon (all pol)
    outgoing: electron (all spin), photon (all pol)

julia> println(proc)
one-photon Compton scattering
```

szabo137 (commented):
There are some suggestions from my side, which we should discuss. I change the status of this PR to draft for the discussion, such that it is not accidentally merged.

docs/Project.toml
AntonReinhard: To use `using QEDbase` in the jldoctests. Without this, I'd have to use `QEDbase.Incoming()` everywhere.

AntonReinhard:
@szabo137 Please have a look again. I reworked the printing and implemented the `MIME...` overloads only for those that need multi-line printing, otherwise just the two-argument `show()`. That means that I should probably rework the printing for the QEDbase types (`Incoming()` etc.) in the same way, as I have done in the `QEDbase_patch.jl` now.

I've also removed most of the jldoctests and only kept the ones that I think add actual value for the documentation, showing an example construction of a `ParticleStateful` and a `PhaseSpacePoint`.

szabo137 (approved):
Looks good to me! Thanks again for this, I like the look and feel now :-)

Issue #58, opened 2024-05-16
Enhancement of phase space point


szabo137:
The current implementation of `ParticleStateful` and `PhaseSpacePoint` could be improved. Here we collect some ideas: 

## particle stateful
- [x] expose the types of the fields as type-parameters
- [x] remove spin information from particle stateful (will be added to the process interface)

## phase space point

- [x] Think about saving only the momenta and wrapping the particle stateful type only on access,
- [x] make `generate_phase_space` private and all it in a constructor of `PhaseSpacePoint`,
- [x] add construction of phase space points from coordinates, maybe using the functions `_generate_incoming_momenta` and `_generate_outgoing_momenta` from the process interface.  
- [x] add abbreviations for `momentum.(psp.in_particles)` and `momentum.(psp.out_particles`

AntonReinhard:
Fully solved with #63 and #68

PR #63, opened 2024-05-19
Proposal for PhaseSpacePoint improvement

AntonReinhard:
Proposal for a rework of the PhaseSpacePoint, according to Issue #58

I did a lot of type magic with recursive variadic templates to find out type information and get functions perfectly type stable. Constructing phase space points is now always type stable and takes a maximum of ~11ns for me when constructing from momenta.

Some of the tests are failing for now because some of the interfaces currently don't expect tuples from the PSP implementation. It might make sense to fix this in #59


AntonReinhard:
I just noticed I'm still missing the new implementation of `spin(...)` / `polarization(...)` of a PSP. I'll add this later

szabo137:
@AntonReinhard Seems like something broke with this one. Maybe it is better to wait with this one until #59 is merged. After that, a rebase of this one seems easy.

AntonReinhard:
Should be building fine for now @szabo137 

I'm not sure if it might actually be a good idea to merge this before #59 instead of the other way around, since there are slight changes in how the PSP behaves. For example, I added a `momenta(::PSP, ::DIR)` function to get an `SVector` of the momenta, because `momentum.(psp.in_particles)` now returns a tuple and not a vector. It doesn't do any copies for this and only takes about 5ns on my machine for a 2-length vector but it might still not be the best approach.

Also, I noticed that there is no public interface for a process' spins/polarizations, which means it's now impossible to get a particle's spin.

AntonReinhard:
I have tentatively merged #59 into this, but keeping it local for now. I'm playing around with it now and am noticing that the broadcast in `momenta(...)` prevents everything from working on the GPU (both ROCm and CUDA).

AntonReinhard:
@szabo137 Should be ready to merge now

test/phase_spaces.jl
```diff
@@ -162,17 +141,21 @@ end
...
     end
 
     @testset "Generation" begin
-        test_psp = generate_phase_space(
-            process, model, phasespace_def, [in_el_mom, in_ph_mom], [out_el_mom, out_ph_mom]
+        test_psp = PhaseSpacePoint(
+            process,
+            model,
+            phasespace_def,
+            SVector(in_el_mom, in_ph_mom),
```
szabo137, line 157: Is there a reason for using `SVectors` here?

szabo137 (commented):
Excellent work! Only some minor comments.

test/phase_spaces.jl
AntonReinhard: Not specifically, it just has to be a `AbstractVector`, so it can't be a Tuple currently. Unless we want to change the interface to accept `NTuples` of `SFourMomentum` maybe.

test/cross_sections.jl
```diff
@@ -13,8 +13,8 @@ TESTPSDEF = TestImplementation.TestPhasespaceDef()
 @testset "($N_INCOMING,$N_OUTGOING)" for (N_INCOMING, N_OUTGOING) in Iterators.product(
     (1, rand(RNG, 2:8)), (1, rand(RNG, 2:8))
 )
-    INCOMING_PARTICLES = rand(RNG, TestImplementation.PARTICLE_SET, N_INCOMING)
-    OUTGOING_PARTICLES = rand(RNG, TestImplementation.PARTICLE_SET, N_OUTGOING)
+    INCOMING_PARTICLES = Tuple(rand(RNG, TestImplementation.PARTICLE_SET, N_INCOMING))
```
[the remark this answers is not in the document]
AntonReinhard: That would require a lot more changes than the few `Tuple(...)` wrappers needed now... I don't think it makes sense to do that.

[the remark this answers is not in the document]
szabo137: Fair enough. I think this is addressed in #67.

test/phase_spaces.jl
szabo137: Same as above, I think this is addressed in #67

szabo137 (approved):
LGTM

Issue #64, opened 2024-05-21
cleanup


szabo137:
Caused by the rapid development, there is some clean-up operation needed. This issue should collect tasks for this. 

- [ ] put `includes` from nested folder structure to `QEDprocesses.jl`
- [ ] separate larger source files in reasonable and functionally separate files (e.g. `phase_space.jl` could easily be a folder with `construction.jl`, `checks.jl`, `particle_stateful.jl`, etc. )
- [ ] reorganize the tests in a similar way
- [ ] move `momentum_generation.jl` to the interfaces

PR #69, opened 2024-05-24
GPU support

AntonReinhard:
This PR adds GPU support. It adds tests that run conditionally on AMDGPU.jl and/or CUDA.jl, depending on whether they are functional on the machine we're running on.

This PR depends on the fix https://github.com/QEDjl-project/QEDbase.jl/pull/64 and is rebased to #68.

Left to do:
- [x] Fix the Compton `_total_probability` function which currently does not work on GPU because of `quadgk` -> opened issue https://github.com/QEDjl-project/QEDprocesses.jl/issues/119
- [x] Add remaining tests for the PSP interface on GPU
- [x] Potentially add testing with KernelAbstractions.jl (does that support broadcasting?), oneAPI.jl and Metal.jl -> Added oneAPI and Metal, KernelAbstractions would make more sense for actual kernels which we don't have yet
- [x] Automatic testing in CI with GPU capable runners
- [x] Add version checks for the GPU tests, since not all libraries work on all versions and there's nothing really we can do about that other than disabling the tests on those versions -> the checks when deciding what GPU tests to run should be enough for now
- [x] Use PkgExtensions to only optionally load the GPU libraries, since this takes a long time and should only be done when specifically requested.

### DIFF
--- a/test/gpu/process_interface.jl
+++ b/test/gpu/process_interface.jl
@@ -29,126 +29,146 @@ PROC_DEF_TUPLES = [
 RNG = Random.MersenneTwister(573)
 
 @testset "Testing with $GPU_MODULE" for (GPU_MODULE, VECTOR_TYPE) in GPUS
-    @testset "$proc $model $ps_def" for (proc, model, ps_def) in PROC_DEF_TUPLES
-        N = 100
-
-        @info "Testing $proc $model $ps_def"
-        flush(stdout)
-
-        psps = [
-            PhaseSpacePoint(
-                proc, model, ps_def, _rand_coordinates(RNG, proc, model, ps_def)...
-            ) for _ in 1:N
-        ]
-        procs = [proc for _ in 1:N]
-
-        gpupsps = VECTOR_TYPE(psps)
-        gpuprocs = VECTOR_TYPE(procs)
-
-        @testset "PSP interface" begin
-            in_moms_gpu = Vector(momenta.(gpupsps, Incoming()))
-            out_moms_gpu = Vector(momenta.(gpupsps, Outgoing()))
-            in_moms = momenta.(psps, Incoming())
-            out_moms = momenta.(psps, Outgoing())
-
-            @test getindex.(in_moms_gpu, Ref(1)) == getindex.(in_moms, Ref(1))
-            @test getindex.(in_moms_gpu, Ref(2)) == getindex.(in_moms, Ref(2))
-            @test getindex.(out_moms_gpu, Ref(1)) == getindex.(out_moms, Ref(1))
-            @test getindex.(out_moms_gpu, Ref(2)) == getindex.(out_moms, Ref(2))
+    @testset "Float type $FLOAT_T" for FLOAT_T in GPU_FLOAT_TYPES[GPU_MODULE]
+        if FLOAT_T != Float64
+            @warn "Skipping tests for $FLOAT_T which is currently not supported!"
+            continue
         end
 
-        @testset "Private Process Functions" begin
-            @test all(
-                isapprox.(
-                    Vector(QEDbase._averaging_norm.(gpuprocs)),
-                    QEDbase._averaging_norm.(procs),
-                ),
-            )
-        end
+        @testset "$proc $model $ps_def" for (proc, model, ps_def) in PROC_DEF_TUPLES
+            N = 100
 
-        @testset "Public Process Functions" begin
-            @test Vector(incoming_particles.(gpuprocs)) == incoming_particles.(procs)
-            @test Vector(outgoing_particles.(gpuprocs)) == outgoing_particles.(procs)
+            @info "Testing $proc $model $ps_def ($FLOAT_T)"
+            flush(stdout)
 
-            @test Vector(particles.(gpuprocs, Incoming())) == particles.(procs, Incoming())
-            @test Vector(particles.(gpuprocs, Outgoing())) == particles.(procs, Outgoing())
+            psps = [
+                PhaseSpacePoint(
+                    proc,
+                    model,
+                    ps_def,
+                    _rand_coordinates(RNG, proc, model, ps_def, FLOAT_T)...,
+                ) for _ in 1:N
+            ]
+            procs = [proc for _ in 1:N]
 
-            @test Vector(number_incoming_particles.(gpuprocs)) ==
-                number_incoming_particles.(procs)
-            @test Vector(number_outgoing_particles.(gpuprocs)) ==
-                number_outgoing_particles.(procs)
+            gpupsps = VECTOR_TYPE(psps)
+            gpuprocs = VECTOR_TYPE(procs)
 
-            @test Vector(number_particles.(gpuprocs, Incoming())) ==
-                number_particles.(procs, Incoming())
-            @test Vector(number_particles.(gpuprocs, Outgoing())) ==
-                number_particles.(procs, Outgoing())
+            @testset "PSP interface" begin
+                in_moms_gpu = Vector(momenta.(gpupsps, Incoming()))
+                out_moms_gpu = Vector(momenta.(gpupsps, Outgoing()))
+                in_moms = momenta.(psps, Incoming())
+                out_moms = momenta.(psps, Outgoing())
 
-            @test Vector(QEDbase.in_phase_space_dimension.(gpuprocs, model)) ==
-                QEDbase.in_phase_space_dimension.(procs, model)
-            @test Vector(QEDbase.out_phase_space_dimension.(gpuprocs, model)) ==
-                QEDbase.out_phase_space_dimension.(procs, model)
-        end
+                @test eltype(eltype(eltype(in_moms_gpu))) == FLOAT_T
+                @test eltype(eltype(eltype(out_moms_gpu))) == FLOAT_T
+                @test eltype(eltype(eltype(in_moms))) == FLOAT_T
+                @test eltype(eltype(eltype(out_moms))) == FLOAT_T
 
-        @testset "Private PSP/Process Interface" begin
-            @test all(
-                isapprox.(
-                    Vector(QEDbase._incident_flux.(gpupsps)), QEDbase._incident_flux.(psps)
-                ),
-            )
+                @test getindex.(in_moms_gpu, Ref(1)) == getindex.(in_moms, Ref(1))
+                @test getindex.(in_moms_gpu, Ref(2)) == getindex.(in_moms, Ref(2))
+                @test getindex.(out_moms_gpu, Ref(1)) == getindex.(out_moms, Ref(1))
+                @test getindex.(out_moms_gpu, Ref(2)) == getindex.(out_moms, Ref(2))
+            end
 
-            @test all(
-                tuple_isapprox.(
-                    Vector(QEDbase._matrix_element.(gpupsps)),
-                    QEDbase._matrix_element.(psps);
-                    rtol=sqrt(eps(Float64)),
-                ),
-            )
+            @testset "Private Process Functions" begin
+                @test all(
+                    isapprox.(
+                        Vector(QEDbase._averaging_norm.(gpuprocs)),
+                        QEDbase._averaging_norm.(procs),
+                    ),
+                )
+            end
 
-            @test Vector(QEDbase._is_in_phasespace.(gpupsps)) ==
-                QEDbase._is_in_phasespace.(psps)
+            @testset "Public Process Functions" begin
+                @test Vector(incoming_particles.(gpuprocs)) == incoming_particles.(procs)
+                @test Vector(outgoing_particles.(gpuprocs)) == outgoing_particles.(procs)
 
-            @test all(
-                isapprox.(
-                    Vector(QEDbase._phase_space_factor.(gpupsps)),
-                    QEDbase._phase_space_factor.(psps),
-                ),
-            )
+                @test Vector(particles.(gpuprocs, Incoming())) ==
+                    particles.(procs, Incoming())
+                @test Vector(particles.(gpuprocs, Outgoing())) ==
+                    particles.(procs, Outgoing())
 
-            # this currently throws an exception because QuadGK does not work on the GPU
-            @test all(
-                isapprox.(
-                    Vector(QEDprocesses._total_probability.(gpupsps)),
-                    QEDprocesses._total_probability.(psps),
-                ),
-            ) broken = true
-        end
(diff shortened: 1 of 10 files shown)


src/gpu.jl
```diff
@@ -0,0 +1,27 @@
...
+    n = length(input_vector)
+    T = eltype(QEDprocesses._eltype_from_psp(input_vector[1]))
+    # copy to gpu
+    roc_input = ROCVector(input_vector)
+
+    # reserve output vector
+    roc_output = ROCVector{T}(undef, n)
+
+    # call kernel
+    t = (16, 1, 1)
```
SimeonEhrig, line 22: AMD GPU has a wavesize of 32 (RDNA - consumer cards) or 64 (CDNA - HPC accerlator).
Therefore 32 should accelerate the code and be safe without checking the HW.

AntonReinhard: The kernel implementation was temporary and the file therefore deleted. You were right of course though ;)

AntonReinhard:
This builds in the CI with 1.10 and rc, while not executing the GPU tests (because there's no GPU on the runners):
```
┌ Warning: No functional GPUs found for testing!
└ @ Main /builds/hzdr/qedjl-project/QEDprocesses-jl/test/gpu/process_interface.jl:23
```
This is by design. To actually use these tests, we would need runners that actually have GPUs, then they would automatically run the tests too.

However, versions 1.6 - 1.9 fail because of dependency issues. Currently, the GPU tests are just normal tests and in the test environment, only packages that are installed in the project can be loaded. This means that whether or not the tests will be run, the dependencies are in the Project.toml and have to be installed when running any tests at all.

I'm not really sure how to solve this. We could add another file such as `runtests_gpu.jl` running the GPU tests and remove the [extras] dependency to the GPU packages, but we would more or less have to manually do the package installation and testing in the CI. I don't think julia has support for something like this, and especially not back in 1.6.

SimeonEhrig:
> This builds in the CI with 1.10 and rc, while not executing the GPU tests (because there's no GPU on the runners):
> 
> ```
> ┌ Warning: No functional GPUs found for testing!
> └ @ Main /builds/hzdr/qedjl-project/QEDprocesses-jl/test/gpu/process_interface.jl:23
> ```
> 
> This is by design. To actually use these tests, we would need runners that actually have GPUs, then they would automatically run the tests too.
> 
> However, versions 1.6 - 1.9 fail because of dependency issues. Currently, the GPU tests are just normal tests and in the test environment, only packages that are installed in the project can be loaded. This means that whether or not the tests will be run, the dependencies are in the Project.toml and have to be installed when running any tests at all.
> 
> I'm not really sure how to solve this. We could add another file such as `runtests_gpu.jl` running the GPU tests and remove the [extras] dependency to the GPU packages, but we would more or less have to manually do the package installation and testing in the CI. I don't think julia has support for something like this, and especially not back in 1.6.

Can we maybe disable tests and the import command via environment variable, like I did it in this Python project: https://github.com/alpaka-group/bashi/blob/c0b673eb1ecff92bde3c3bb89c277104cdbedde8/tests/test_generate_combination_list.py#L186

CUDA.jl also provides a method to detect GPU's: https://cuda.juliagpu.org/stable/installation/conditional/

AntonReinhard:
> Can we maybe disable tests and the import command via environment variable, like I did it in this Python project: https://github.com/alpaka-group/bashi/blob/c0b673eb1ecff92bde3c3bb89c277104cdbedde8/tests/test_generate_combination_list.py#L186
>
> CUDA.jl also provides a method to detect GPU's: https://cuda.juliagpu.org/stable/installation/conditional/

No, we can't use either of these, because the problem doesn't happen while executing the tests (the tests already are skipped when the libraries are non-functional for whatever reason).
The problem happens when resolving the Project.toml.

AntonReinhard:
> Looks like CUDA.jl and AMDGPU.jl are hard dependencies for the test, or isn't it? If yes, can remove it from the Project.toml and test in the test scripts, if the packages are installed?

At least as far as I'm aware this does not work. I tried having AMDGPU.jl globally installed but not in the test dependencies, and then load it inside the test. But it just says it's not installed.
Maybe this would work if not used through `Pkg.test()`, but then we would more or less start manually implementing Julia's testing framework...

AntonReinhard:
For reference, the part that actually fails is not the `Pkg.instantiate()`. That works because the GPU packages are only weak dependencies.
When running the tests and creating their environment, the error looks like this (from CI with Julia1.9):
```
$ julia --project=. -e 'import Pkg; Pkg.test(; coverage = true)'
     Testing QEDprocesses
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations /usr/local/julia/share/julia/stdlib/v1.9/Pkg/src/Operations.jl:1814
ERROR: Unsatisfiable requirements detected for package LLD_jll [d55e3150]:
 LLD_jll [d55e3150] log:
 ├─possible versions are: 14.0.6 or uninstalled
 └─found to have no compatible versions left with AMDGPU [21141c5a]
   └─AMDGPU [21141c5a] log:
     ├─possible versions are: 0.1.0-1.0.0 or uninstalled
     ├─restricted to versions 1 by QEDprocesses [46de9c38], leaving only versions: 1.0.0 or uninstalled
     │ └─QEDprocesses [46de9c38] log:
     │   ├─possible versions are: 0.2.0 or uninstalled
     │   └─QEDprocesses [46de9c38] is fixed to version 0.2.0
     └─restricted to versions 1 by an explicit requirement, leaving only versions: 1.0.0
```

szabo137:
> For reference, the part that actually fails is not the `Pkg.instantiate()`. That works because the GPU packages are only weak dependencies.
> 
> When running the tests and creating their environment, the error looks like this (from CI with Julia1.9):
> 
> ```
> 
> $ julia --project=. -e 'import Pkg; Pkg.test(; coverage = true)'
> 
>      Testing QEDprocesses
> 
> ┌ Warning: Could not use exact versions of packages in manifest, re-resolving
> 
> └ @ Pkg.Operations /usr/local/julia/share/julia/stdlib/v1.9/Pkg/src/Operations.jl:1814
> 
> ERROR: Unsatisfiable requirements detected for package LLD_jll [d55e3150]:
> 
>  LLD_jll [d55e3150] log:
> 
>  ├─possible versions are: 14.0.6 or uninstalled
> 
>  └─found to have no compatible versions left with AMDGPU [21141c5a]
> 
>    └─AMDGPU [21141c5a] log:
> 
>      ├─possible versions are: 0.1.0-1.0.0 or uninstalled
> 
>      ├─restricted to versions 1 by QEDprocesses [46de9c38], leaving only versions: 1.0.0 or uninstalled
> 
>      │ └─QEDprocesses [46de9c38] log:
> 
>      │   ├─possible versions are: 0.2.0 or uninstalled
> 
>      │   └─QEDprocesses [46de9c38] is fixed to version 0.2.0
> 
>      └─restricted to versions 1 by an explicit requirement, leaving only versions: 1.0.0
> 
> ```

It seems like the compat entry for AMDGPU is too restrictive. Anyway, since AMDGPU itself only supports Julia versions >=1.10, I suggest to drop the support for AMDGPU for all julia versions below 1.10 as well. Therefore, we only need to test versions, which already pass the tests. 
In the end, this is not problematic, because, I think we will drop the support for <1.10 if 1.10 becomes the LTS, i.e. if 1.11 is released.

test/gpu/process_interface.jl
```diff
@@ -0,0 +1,161 @@
...
+    println("Testing with CUDA.jl")
+    push!(GPU_VECTOR_TYPES, CuVector)
+end
+
+if isempty(GPU_VECTOR_TYPES)
+    @warn "No functional GPUs found for testing!"
+    return nothing
+end
+
+PROC_DEF_TUPLES = [
```
szabo137, line 27: Could you iterate over all spin/pol combinations here using `Iterators.product`?

test/gpu/process_interface.jl
```diff
@@ -0,0 +1,161 @@
...
+                ),
+            )
+
+            #= TODO: this is currently broken
+            @test all(
+                isapprox.(
+                    Vector(QEDprocesses._total_probability.(gpupsps)),
+                    QEDprocesses._total_probability.(psps),
+                ),
+            )=#
```
szabo137, line 130: FYI: there is the macro [`@test_broken`](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_broken) to indicate a test is broken. This will then be shown in the test statistics as well.

test/gpu/process_interface.jl
```diff
@@ -0,0 +1,161 @@
...
+            )
+
+            @test all(
+                isapprox.(
+                    Vector(differential_cross_section.(gpupsps)),
+                    differential_cross_section.(psps),
+                ),
+            )
+
+            #= TODO: this is currently broken
```
szabo137, line 155: Same as above, consider using `@test_broken` here.

test/test_implementation/random_coordinates.jl
```diff
@@ -0,0 +1,16 @@
+"""
+Return a tuple of tuples of incoming and outgoing coordinates for a given process, model and ps_def that make up a physical phase space point.
+"""
+function _rand_coordinates(
+    rng::AbstractRNG, ::PROCESS, ::MODEL, ::PS_DEF
+) where {PROCESS<:Compton,MODEL<:PerturbativeQED,PS_DEF<:PhasespaceDefinition}
+    return ((rand(rng, Float64),), (rand(rng, Float64), rand(rng, Float64)))
+end
+
+tuple_isapprox(::Tuple{}, ::Tuple{}; atol=0.0, rtol=eps()) = true
```
szabo137, line 10: This only works if the tuples have the same length, maybe you should assert this here as well.

szabo137 (commented):
Nice work so far, here are some comments from my side.

test/gpu/process_interface.jl
AntonReinhard: Ah right. I was aware of the macro, but I wasn't aware it also handles exceptions. I will use it then

test/test_implementation/random_coordinates.jl
AntonReinhard: Since it's just the negative case where it should fail anyways, I left it out originally, but it might be nicer to have a `return false` instead of a `MethodError` so I added the overloads.

test/gpu/process_interface.jl
AntonReinhard: I can easily do that, the problem just becomes execution time. Testing so many GPU kernels takes a long time, so even with just the 17 total cases it now already takes ~2.5 minutes on my machine. This is fine for now I think, but we might have to reconsider some numbers in the future when the tests get even more extensive.

AntonReinhard:
I opened a discourse thread about our testing problem, but there doesn't really seem to be an answer (yet)
https://discourse.julialang.org/t/testing-gpu-compatability-in-ci/119021

It seems the only real option is to more or less setup GPU tests with PackageExtensions manually in the CI by loading the necessary packages only on julia versions and runners/architectures where they will compile.

AntonReinhard:
Since we have now dropped Julia versions 1.9 and lower, we can properly use package extensions and load AMDGPU and CUDA properly, even when no supported GPU is available. Currently, the tests will simply not run when no GPU is found. So if we get a runner with a GPU and run unit tests on that, it should run the tests.

AntonReinhard:
What do you think how we should proceed with this PR from here @szabo137 ?

szabo137:
> What do you think how we should proceed with this PR from here @szabo137 ?

As discussed offline, we should keep this, at least, as a testing field for the integration of GPU tests. However, maybe we should think about having such a *testing branch* upstream with less actual functionality. Maybe GPU-tests for CuArrays of `SFourMomenta` or PSPs in `QEDcore` would be easier and more convenient. Then this PR here could be used to add the actual GPU tests for `QEDprocesses` after we agreed on an actual workflow.

SimeonEhrig:
> I'm a bit confused why the GPU tests fail. The AMDGPU tests crash violently in the CI, but pass fine on my own AMDGPU. The CUDA tests seem to compute incorrect results, which also doesn't really make sense to me.

It looks like a compiler and not a rutime error. This is strange. The only explanation which I have is, that it cannot correctly detect the GPU maybe some features are not enabled.

Can you reproduce it in a container? Have a look in the generated yaml code to reproduce it locally (search for the job unit_test_julia_amdgpu_1_11): https://gitlab.com/hzdr/qedjl-project/QEDprocesses-jl/-/jobs/8517785749

AntonReinhard:
> Can you reproduce it in a container? Have a look in the generated yaml code to reproduce it locally (search for the job unit_test_julia_amdgpu_1_11): https://gitlab.com/hzdr/qedjl-project/QEDprocesses-jl/-/jobs/8517785749

I'm not sure how to do that, maybe you could show me some time? It still seems strange, because if the problem is in the runner, then shouldn't the GPU tests on QEDbase have failed too?

AntonReinhard:
It seems that AMDGPU.jl had a problem with Float64 somewhere. I removed the Float64 from the tests for now and opened an issue [here](https://github.com/QEDjl-project/QEDprocesses.jl/issues/123).

The second error seems to have been caused by a difference in metric used for `isapprox` between GPU vectors and normal vectors. Probably related to [this](https://github.com/JuliaGPU/GPUArrays.jl/issues/191) in some form.

AntonReinhard:
Update: the CUDA problem was with `isapprox`, but it's a user error. The `_all_onshell` function checked against zero when the mass of the particle is zero, i.e., for photons. In this case `isapprox` is the same as `x == 0`. The arithmetic apparently differs slightly between CPU and GPU, so these didn't always match. It was definitely an error that had to be fixed in any case.

test/gpu/process_interface.jl
```diff
@@ -0,0 +1,174 @@
...
+                            return QEDbase._averaging_norm(T, proc)
+                        end
+
+                        @test all(
+                            isapprox.(Vector(_stable_norm.(gpuprocs)), _stable_norm.(procs))
+                        )
+                    end
+                    wrap(FLOAT_T)
+                end
+            end
```
szabo137, line 80: 🤨🤔🙄☝️😅

Indeed, not pretty, but ok. We just need to remove this after we improve the interface.

szabo137 (approved):
Very nice, and many thanks for your long-standing support on this!

test/gpu/process_interface.jl
AntonReinhard: :innocent:

SimeonEhrig:
Merged :partying_face:

PR #71, opened 2024-05-27
Update Julia version in CI

AntonReinhard:
As the title says, update the Julia version used in the CI to 1.10, and use 1.10 and 1.11 additionally in automated unit tests.


SimeonEhrig:
Because Julia 1.11 is not released yet, there is no container with the tag.

https://hub.docker.com/_/julia

We could test `rc` but I'm not sure, what happen if Julia 1.11 is released. Maybe it will be removed, stuck to 1.11 or moved the 1.12 dev.

AntonReinhard:
> We could test `rc` but I'm not sure, what happen if Julia 1.11 is released. Maybe it will be removed, stuck to 1.11 or moved the 1.12 dev.

That seems like the most reasonable option. I'd suspect that once 1.11 is released, the `latest` tag will be moved to 1.11 and the `rc` tag will be moved to 1.12. In any case we will have to update the CI again anyways when that happens, but at least it should not start to fail using `rc` then.

szabo137 (dismissed):
Looks good to me. Thanks!

SimeonEhrig:
There are warnings in Julia 1.9 and Julia 1.10 which should be fixed a upcoming PR. Also for 1.10 there is a error for pre-compiling. But it looks like, this is not a big problem.

szabo137 (approved):
LGFM. Merge after pipeline succeeds.

PR #72, opened 2024-05-31
[Proposal] add pre-commit hooks

szabo137:
TBW

# TODO

- [ ] describe the init of pre-commit in the README/docs
- [ ] discuss the default config


AntonReinhard:
Didn't know about this tool, just read up on it a bit and it seems really useful. Do we plan on using the https://pre-commit.ci/ bot too? That would cover every contribution, not just those who have the tool installed themselves, and also changes through the web interface, like applying code suggestions.

szabo137:
I like the idea of using the pre-commit bot, but I don't know if we should test it first.

PR #78, opened 2024-06-19
Use base state from core

szabo137:
With this PR, QEDprocesses should use `base_state` and all particles from QEDcore, not from QEDbase. 

This is part of the general restructuring of `QED.jl`, see https://github.com/QEDjl-project/QED.jl/issues/35 for details.


szabo137:
@AntonReinhard ping 😄

AntonReinhard (approved):

PR #79, opened 2024-06-20
bump dev version

szabo137:
Bumps the current version of QEDprocesses to a 0.2-DEV.

Reason: during the current refac, we need the current dev registered to have working integration tests.


AntonReinhard (approved):

szabo137:
Seems like this is not necessary anymore, because #78 went through. Close without merging.

PR #80, opened 2024-06-21
Qualify total_cross_section calls

AntonReinhard:
Since `total_cross_section` moves to QEDbase, we need to qualify the calls here for now in order for integration tests to work.

Once we remove the implementation here, the `QEDbase.` calls will automatically fail and need to be removed as well.


szabo137 (approved):
Looks good.

PR #81, opened 2024-06-21
Bump dev version

szabo137:
This bumps the version of the dev-branch to 0.1.1-DEV. 

This in only for test purposes and should not me released in the general registry. 
It only ticks up the patch version to not break compats upstream.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ authors = [
     "Tom Jungnickel",
     "Anton Reinhard",
 ]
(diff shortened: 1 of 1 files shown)


AntonReinhard (approved):
Looks good, but should we merge #80 before this or does it not matter? I'm not sure

szabo137:
> Looks good, but should we merge #80 before this or does it not matter? I'm not sure

I think it's better to merge #80 first (already merged) because we want the integration test to pass.

PR #94, opened 2024-08-13
Fix for multiplicity change in QEDbase

AntonReinhard:
See https://github.com/QEDjl-project/QEDbase.jl/pull/112

### DIFF
--- a/src/processes/one_photon_compton/perturbative/cross_section.jl
+++ b/src/processes/one_photon_compton/perturbative/cross_section.jl
@@ -21,8 +21,7 @@ end
     We average over the initial spins and pols, and sum over final.
 """
 function QEDbase._averaging_norm(proc::Compton)
-    normalizations = multiplicity.(_in_spin_and_pol(proc))
-    return inv(prod(normalizations))
+    return inv(incoming_multiplicity(proc))
 end
 
 @inline function _all_onshell(psp::PhaseSpacePoint{<:Compton})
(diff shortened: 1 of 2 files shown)


AntonReinhard:
@SimeonEhrig I think there's a minor problem with the CI here: The docs fail because the `CI_UNIT_PKG_URL_QEDbase` is only used by the gitlab ci and not the github actions file. Because of that, the project can't be precompiled and the docs can't be built. 

It's not too important because the update in QEDbase will be merged first and afterwards this problem will go away, but it would still be nice if it worked. Could that be easily added?

szabo137 (approved):

SimeonEhrig:
> @SimeonEhrig I think there's a minor problem with the CI here: The docs fail because the `CI_UNIT_PKG_URL_QEDbase` is only used by the gitlab ci and not the github actions file. Because of that, the project can't be precompiled and the docs can't be built.
> 
> It's not too important because the update in QEDbase will be merged first and afterwards this problem will go away, but it would still be nice if it worked. Could that be easily added?

Do I understand it right that the problem is that `incoming_spin_pols` is defined in the dev branch of QEDbase and the documentation builds with the latest release?

AntonReinhard:
> Do I understand it right that the problem is that incoming_spin_pols is defined in the dev branch of QEDbase and the documentation builds with the latest release?

`incoming_spin_pols` is defined in the feature branch that I linked with the `CI_UNIT_PKG_URL_QEDbase: <commit>` in the last commit message. But the documentation does build with the latest release which will be a problem because it should be the latest dev. The latest release will not get the function definition until after QEDbase is released again, and that should not be necessary.

AntonReinhard:
Ok yes, the feature in QEDbase has been merged but the documentation building still fails here because it uses the released version of QEDbase, not its `dev` branch.

We could change the `docs/make.jl` script to always depend on dev versions of QED packages, that might be the easiest way to circumvent this problem. It shouldn't get in the way with compatibility testing because that's what the gitlab CI is doing anyways. What do you think @SimeonEhrig ?

SimeonEhrig:
Please rebase. #95 should solve the CI build doc problem.

SimeonEhrig (approved):

PR #96, opened 2024-08-21
GenericQEDProcess

AntonReinhard:
Adds a generic QED process, implementing the basic `AbstractProcessDefinition` interface and allowing any combination of particles and spins/pols for them.
The constructor validates that the given spins and polarizations fit to the particles (fermion -> spin, boson -> polarization).

The actual computational functions are as of yet unimplemented and will later be added using the QEDfeynman_diagrams project, once that is public.

Therefore, for now the main use of this type is for testing purposes of other functions.

### DIFF
--- a/src/processes/generic_process/process.jl
+++ b/src/processes/generic_process/process.jl
@@ -1,15 +1,30 @@
 """
-    GenericQEDProcess <: AbstractProcessDefinition
+    ScatteringProcess <: AbstractProcessDefinition
 
-An implementation of the `AbstractProcessDefinition` interface for a generic particle process in QED
-with any number of incoming and outgoing particles, and any combination of spins or polarizations for the particles set.
+Generic implementation for scattering processes of arbitrary particles. Currently, only calculations in combination with `PerturbativeQED` are supported. 
+However, this is supposed to describe scattering processes with any number of incoming and outgoing particles, and any combination of spins or polarizations for the particles.
 
-The [`isphysical`](@ref) function can be used to check whether the process is possible in QED.
+The [`isphysical`](@ref) function can be used to check whether the process is possible in perturbative QED.
 
-!!! note
+!!! warning  
     The computation of cross sections and probabilities is currently unimplemented.
+
+## Constructors
+
+    ScatteringProcess(
+        in_particles::Tuple{AbstractParticleType},
+        out_particles::Tuple{AbstractParticleType},
+        [in_sp::Tuple{AbstractSpinOrPolarization},
+        out_sp::Tuple{AbstractSpinOrPolarization}]
+    )
+
+Constructor for a ScatteringProcess with the given incoming and outgoing particles and their respective spins and pols.
+The constructor asserts that the particles are compatible with their respective spins and polarizations. If the assertion fails, an 
+`InvalidInputError` is thrown.
+
+The `in_sp` and `out_sp` parameters can be omitted in which case all spins and polarizations will be set to `AllSpin` and `AllPol` for every fermion and boson, respectively.
 """
-struct GenericQEDProcess{INT,OUTT,INSP,OUTSP} <:
+struct ScatteringProcess{INT,OUTT,INSP,OUTSP} <:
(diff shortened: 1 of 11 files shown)


src/processes/generic_process/utility.jl
```diff
@@ -0,0 +1,45 @@
+_assert_particle_type_tuple(::Tuple{}) = nothing
```
szabo137, line 1: Since recursive multiple dispatch is hard to infer for a reader, adding some comments here would be convenient.

src/processes/generic_process/utility.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+function _assert_particle_type_tuple(t::Any)
+    throw(
+        InvalidInputError(
+            "invalid input, provide a tuple of AbstractParticleTypes to construct a GenericQEDProcess",
+        ),
+    )
+end
+
+_assert_spin_pol_particle_compatability(::Tuple{}, ::Tuple{}) = nothing
```
szabo137, line 13: Same here, some comments would help understand what is asserted.

src/processes/generic_process/utility.jl
```diff
@@ -0,0 +1,45 @@
...
+        throw(
+            InvalidInputError(
+                "particle \"$(particles[1])\" is a boson and should have a polarization, but has \"$(spin_pols[1])\"",
+            ),
+        )
+    end
+    return _assert_spin_pol_particle_compatability(particles[2:end], spin_pols[2:end])
+end
+
+# this should move to QEDbase as part of the interface, see https://github.com/QEDjl-project/QEDbase.jl/issues/114
```
szabo137, line 42: For now, this is fine, because there is only internal use. However, for external users, I have some suggestions I will post below the referred issue.

test/processes/generic_process/process.jl
```diff
@@ -0,0 +1,157 @@
+using QEDprocesses
+using Random
+using QEDcore
+
+include("groundtruths.jl")
+
+const RNG = MersenneTwister(77697185)
+
+PARTICLES = [Electron(), Positron(), Photon()]
+POLS = [PolX(), PolY(), AllPol()]
```
szabo137, line 10: Should the `SyncedSpin`/`SyncedPolarization` functionality tested here as well? Or is this something for the cross-section tests?

szabo137 (commented):
Very good, thanks for providing this! I like the clean structure of the code. There are just some minor concerns and some naming things from my side.

src/processes/generic_process/process.jl
```diff
@@ -0,0 +1,127 @@
...
+end
+function QEDbase.incoming_spin_pols(proc::GenericQEDProcess)
+    return proc.incoming_spin_pols
+end
+function QEDbase.outgoing_spin_pols(proc::GenericQEDProcess)
+    return proc.outgoing_spin_pols
+end
+
+"""
+    isphysical(proc::GenericQEDProcess)
```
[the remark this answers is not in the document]
AntonReinhard: I have moved it to utils and overloaded it with AbstractProcessDefinition and PerturbativeQED. The extension of the particle interface for electron_numbers sounds like a good idea and I will open an issue for this in QEDbase.

However, I don't think this should be checked in the construction of a process. I think the fermion->spin and boson->pol should be checked because if those don't fit together, things will break from an implementation standpoint. On the other hand, having a representation of a physically impossible process doesn't really hurt the implementation, and sensible return values for the functions exist (for example probabilities are 0).

test/processes/generic_process/process.jl
AntonReinhard: There's not really any reason not to test it here too, so I added them.

src/processes/generic_process/utility.jl
AntonReinhard: I was actually able to remove this one entirely and simplify the _assert_spin_pol_particle_compatability as well. I added some more comments and hope it is fine now.

szabo137 (approved):

PR #105, opened 2024-10-25
bumped compat entry for QEDbase

szabo137:
CI_INTG_PKG_URL_QEDcore=https://github.com/szabo137/QEDcore.jl#bump-QEDbase-compat

- [ ] check if https://github.com/QEDjl-project/QEDcore.jl/pull/78 is merged

PR #106, opened 2024-10-28
Fix CI scripts, update compats

AntonReinhard:
Update CI scripts to the new version, same as QEDbase and QEDcore already have.
Also update compat section here as well so the CI actually works.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,2 +1,1 @@
 style = "blue"
-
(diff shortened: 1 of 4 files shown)


AntonReinhard:
This will make #105 and #104 redundant, but we need to update the compat together with the update of the CI scripts to be able to merge.

SimeonEhrig (approved):

PR #115, opened 2025-02-06
updated readme

szabo137:


SimeonEhrig (approved):

SimeonEhrig:
@szabo137 After merging, I saw that the badges are not in a row. They are column wise. I think this was not intended?

PR #122, opened 2025-04-14
Add runic

szabo137:
This is a proposal to use [runic](https://github.com/fredrikekre/Runic.jl?tab=readme-ov-file#checking-formatting) for formatting instead of `JuliaFormatter`. 

# TODO

- [x] Remove `JuliaFormatter` from the repository
- [x] Add [`runic-action`](https://github.com/fredrikekre/runic-action) to the gh-action workflows (replacing the old formatting workflow)
- [x] Replace the code style badge with the one from Runic.


AntonReinhard:
This needs to be rebased now, then we can merge it.

.DS_Store
AntonReinhard, line 1: This was added by accident? Should probably not be here

.pre-commit-config.yaml
AntonReinhard, line 1: Do we want to add this in the other repositories as well?

.DS_Store
szabo137: removed in c01527e

.pre-commit-config.yaml
szabo137: yep, but not in this PR... removed in c01527e

AntonReinhard (approved):

PR #124, opened 2025-04-15
Run fewer GPU tests in CI

AntonReinhard:
Adds a testing util function `LARGE_TESTS()` that returns true if larger tests should be run.

If the environment variable `LARGE_TESTS` is explicitly set, it will run them ("1") or skip them ("0"). If it's not set, they will by default be run locally and skipped in the CI, i.e., when `CI` is set.

Currently, this just disables some of the spin/pol combinations tested in the GPU tests, but could be used for more future tests as well. Currently, there is no way to activate the LARGE_TESTS in the CI for a specific run, because we cannot manually set environment variables in the scripts.

### DIFF
--- a/test/gpu/process_interface.jl
+++ b/test/gpu/process_interface.jl
@@ -28,6 +28,11 @@ PROC_DEF_TUPLES = [
 
 RNG = Random.MersenneTwister(573)
 
+if !LARGE_TESTS()
+    @info "Skipping large tests...\nEnable them explicitly with an environment variable LARGE_TESTS=1"
+    PROC_DEF_TUPLES = PROC_DEF_TUPLES[1:5]
+end
+
 @testset "Testing with $GPU_MODULE" for (GPU_MODULE, VECTOR_TYPE) in GPUS
     @testset "Float type $FLOAT_T" for FLOAT_T in GPU_FLOAT_TYPES[GPU_MODULE]
         @testset "$proc $model $psl" for (proc, model, psl) in PROC_DEF_TUPLES
--- a/test/gpu/runtests.jl
+++ b/test/gpu/runtests.jl
@@ -96,6 +96,7 @@ if metal_tests
 end
 
 include("../test_implementation/random_coordinates.jl")
+include("../utils.jl")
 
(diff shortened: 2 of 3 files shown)


test/gpu/process_interface.jl
```diff
@@ -28,6 +28,11 @@ PROC_DEF_TUPLES = [
 
 RNG = Random.MersenneTwister(573)
 
+if !LARGE_TESTS()
+    @info "Skipping large tests...\nEnable them explicitly with an environment variable LARGE_TESTS=1"
+    PROC_DEF_TUPLES = PROC_DEF_TUPLES[1:5]
```
SimeonEhrig, line 33: What does the line actually do? Does it remove test cases? And if so, is it okay that test parameters are removed without knowing the actual test parameters?

AntonReinhard: The line checks for environment variables, `LARGE_TESTS` enables or disables specifically, otherwise `CI` disables them. This means that locally they will run by default.
What this removes are some of the spin/pol combinations that are tested. Everything really relevant is tested in the first few, i.e. every particle gets every spin/pol once, just some of the combinations will be skipped. It was just meant to save some time, right now skipping these saves probably around 10min per run in the CI.
For now I think it's fine like this and in the future it might be nice to be able to manually enable `LARGE_TESTS` for the CI or have them run for releases only.

szabo137 (approved):
LGTM, however, maybe it's better to choose the tested combination randomly instead of consecutively. Best solution might be to use some [all-pairs](https://en.m.wikipedia.org/wiki/All-pairs_testing) sparsity pattern.

AntonReinhard:
> LGTM, however, maybe it's better to choose the tested combination randomly instead of consecutively. Best solution might be to use some [all-pairs](https://en.m.wikipedia.org/wiki/All-pairs_testing) sparsity pattern.

I'm (still) not a fan of using RNG to choose the test cases. All-pars testing would be nice but also probably overkill for right now.
If nobody has concerns I'd just merge it like this for now.

SimeonEhrig (approved):

PR #126, opened 2025-04-23
Fix new GPU test failure

AntonReinhard:
My best guess is that one of the random momenta hits an unstable spot yielding bad precision, since it only happens so rarely. Maybe we'll have to add a `singularity_proximity` measure or similar to check beforehand if it's likely that a psp will be unstable.


AntonReinhard:
This is not really a fix and more like a cover. I'm currently not sure how we want to go about fixing this sort of problem, because I don't quite understand where it's coming from in the first place...

szabo137:
As discussed offline, this should be fine for now. Could up please open/update an issue, indicating that this might need a more sophisticated solution.

szabo137 (approved):

PR #129, opened 2025-05-15
Add qed-fd.jl to the dependencies and implement matrix element

AntonReinhard:
Implements the process interface for the `ScatteringProcess` using QEDFeynmanDiagrams.jl.

Progress:
- [x] `_incident_flux(psp::InPhaseSpacePoint{ScatteringProcess,PerturbativeQED})`
- [x] `_matrix_element(psp::PhaseSpacePoint{ScatteringProcess,PerturbativeQED})`
- [x] `_averaging_norm(::Type{T}, proc::ScatteringProcess)`
- [x] `_is_in_phasespace(psp::PhaseSpacePoint{ScatteringProcess,PerturbativeQED})`
- [x] `_phase_space_factor(psp::PhaseSpacePoint{ScatteringProcess,PerturbativeQED,PSL})`
  
I'm unsure about the implementation for `_incident_flux` and `_phase_space_factor`.


szabo137:
As discussed offline, there is no generic implementation of `_incident_flux` for arbitrary N to M processes. We could only provide generic implementations for 2 to M processes or 1+n gamma to M processes. 
Would it be reasonable to check for these two cases upon construction of the `ScatteringProcess` and raise a warning otherwise?

AntonReinhard:
> As discussed offline, there is no generic implementation of `_incident_flux` for arbitrary N to M processes. We could only provide generic implementations for 2 to M processes or 1+n gamma to M processes. Would it be reasonable to check for these two cases upon construction of the `ScatteringProcess` and raise a warning otherwise?

I think a warning would make sense, yes, since some functionality would still be available.
I will do that and then mark this ready.

src/processes/generic_process/perturbative/cross_section.jl
```diff
@@ -1,20 +1,64 @@
...
+    g = graph(proc)
+    # TODO: there seems to be a bug in 1.12 with ComputableDAGs.jl. Enable this again once it's fixed
+    #optimize_to_fixpoint!(ReductionOptimizer(), g)
+    return get_compute_function(g, proc, cpu_st(), @__MODULE__; closures_size = 1000, concrete_input_type = PSP)
+end
 
-function QEDbase._is_in_phasespace(psp::PhaseSpacePoint{<:ScatteringProcess,PerturbativeQED}) end
+function QEDbase._matrix_element(psp::PhaseSpacePoint{PROC, PerturbativeQED}) where {PROC <: ScatteringProcess}
+    mat_el_func = _mat_el_func(process(psp), typeof(psp))
+    return sqrt(mat_el_func(psp))
```
szabo137, line 37: I assume the `mat_el_func` already returns the square of the matrix element. Therefore, it might be more convenient to add it directly to `QEDbase._matrix_element_square` and keep this for testing purposes.

Furthermore, is this a scalar square root? I am asking, because the sum over the spins and pols needs to be done *after* mod-squaring the matrix elements for every spin-pol combination.

src/QEDprocesses.jl
```diff
@@ -15,8 +15,9 @@ export ComptonSphericalLayout
 # generic scattering process
 export ScatteringProcess, isphysical
 
-using QEDbase
-using QEDcore
+using Reexport
+@reexport using QEDbase
+@reexport using QEDcore
```
szabo137: Why do you reexport `QEDbase` and `QEDcore`? I don't know if this is a good idea, because the packages should be independent and replaceable 🤔

src/processes/generic_process/perturbative/cross_section.jl
```diff
@@ -1,20 +1,64 @@
...
+    return _all_onshell(particles(psp, Incoming()), momenta(psp, Incoming())) &&
+        _all_onshell(particles(psp, Outgoing()), momenta(psp, Outgoing()))
+end
+
+function QEDbase._incident_flux(psp::InPhaseSpacePoint{PROC, PerturbativeQED}) where {PROC <: ScatteringProcess}
+    if incoming_particles(PROC) > 2
+        throw("_incident_flux is unimplemented for general scattering processes with more than 2 incoming particles")
+    end
+
+    # TODO
```
szabo137: The correct formula for 2->N processes is 

```julia
sqrt((pa*pb)^2 - (ma*mb)^2)
```

szabo137 (commented):
Just added some comments. Would you mind adding some unit tests, at least checking if the implementation of the differential cross section interface works as expected?

src/processes/generic_process/perturbative/cross_section.jl
AntonReinhard: > I assume the mat_el_func already returns the square of the matrix element. Therefore, it might be more convenient to add it directly to QEDbase._matrix_element_square and keep this for testing purposes.

True, I can just implement both of these.

> Furthermore, is this a scalar square root? I am asking, because the sum over the spins and pols needs to be done after mod-squaring the matrix elements for every spin-pol combination.

Yes, `mat_el_func` essentially does `sum(abs2(x))` over the spin/pol combinations, so this is a scalar square root.

src/QEDprocesses.jl
AntonReinhard: You're right, I'll remove it.

AntonReinhard:
> Would you mind adding some unit tests, at least checking if the implementation of the differential cross section interface works as expected?

Right, I kinda forgot about that. I could try to test against Klein-Nishina as well.

src/processes/generic_process/perturbative/cross_section.jl
AntonReinhard: Added in 0c9e78c401c9604dce63d5b044d5e3964ece4143, can you check if I did it correctly

src/QEDprocesses.jl
AntonReinhard: Done in 8680a79775261139f4e7aba305294b44815b08b4

src/processes/generic_process/perturbative/cross_section.jl
AntonReinhard: Added in 10e3e80129b6956b7bce21a89ad92e18679a8811

src/processes/generic_process/perturbative/cross_section.jl
szabo137: Argh, Github does not support math rendering. I added the original comment and the respective Julia code. 

But your implementation seems right!

src/processes/generic_process/perturbative/cross_section.jl
```diff
@@ -1,20 +1,78 @@
...
+            )
+        )
+        return false
+    end
+    return _all_onshell(particles(psp, Incoming()), momenta(psp, Incoming())) &&
+        _all_onshell(particles(psp, Outgoing()), momenta(psp, Outgoing()))
+end
+
+function QEDbase._incident_flux(psp::InPhaseSpacePoint{PROC, PerturbativeQED}) where {PROC <: ScatteringProcess}
+    proc = PROC()
```
szabo137:
```suggestion
    proc = process(psp)
```

just nitpicking

AntonReinhard:
I added tests, but the pipelines will fail until https://github.com/QEDjl-project/QEDFeynmanDiagrams.jl/pull/53 is merged

src/processes/generic_process/perturbative/cross_section.jl
AntonReinhard: It was actually not working like this, so good nitpick ;)

szabo137 (approved):
LGTM!

PR #131, opened 2025-06-16
Lower test precision

AntonReinhard:


szabo137 (dismissed):
LGTM!

szabo137:
There is something wrong with the CUDA tests. There is a problem with the public API for PSPs.

SimeonEhrig:
Oh no, not again :-(
https://gitlab.com/hzdr/qedjl-project/QEDprocesses-jl/-/jobs/10362481707

AntonReinhard:
I'm somewhat blindly trying here, since I can't reproduce this locally... Hopefully I got it now

szabo137 (approved):
Not the best solution, but currently, there is no better one 🤷‍♂️

PR #133, opened 2025-06-27
removed constants in favor of QEDcore#121

szabo137:
Fixes integration tests for https://github.com/QEDjl-project/QEDcore.jl/pull/121

PR #134, opened 2025-06-30
Fix dcs metal

szabo137:
Calculating differential cross sections failed when broadcasting over an `MtlVector` of phase space points. The construction of the tuple of matrix elements for each spin-pol combination causes this. It seems like Julia does not unroll it properly, if the number of elements is larger than 10 (which is the case for the fully unpolarized Compton). Therefore, a replacement of the tuple construction by a version using `StaticVector`, which is fully unrolled, fixes the problem. 

consider for merging after #133


AntonReinhard (dismissed):

AntonReinhard:
Oh, looks like we need to replace some `tuple_isapprox` with an `svector_isapprox`

AntonReinhard (approved):

szabo137:
> Oh, looks like we need to replace some `tuple_isapprox` with an `svector_isapprox`

Indeed, but since this is not exported and only used in the tests, I suggest leaving it like that and changing it later if necessary 🤷

PR #136, opened 2025-07-25
Fix float type in _incident_flux of ScatteringProcess

AntonReinhard:
Previously the function promoted to Float64 by default


szabo137 (approved):
Pretty good catch, thanks for fixing 😸

PR #137, opened 2025-07-29
Fix onshell check precisions

AntonReinhard:
Also swaps the implementation to work on `ParticleStatefuls` everywhere


szabo137 (approved):
LGTM 😸

PR #139, opened 2025-10-15
Release Version 0.5.0 to main

AntonReinhard:
Do not squash, use simple merge commit!

This has to be a breaking release because the exported constants (`ALPHA`, `ALPHA_SQUARE`, etc.) were removed in #133  and moved to QEDcore.


szabo137 (approved):
LGTM!

PR #141, opened 2025-10-29
Implement the new square sum matrix element interface from QEDbase

AntonReinhard:
See https://github.com/QEDjl-project/QEDbase.jl/pull/176
This is useful for a fix to the problem in https://github.com/QEDjl-project/QEDevents.jl/pull/70

### DIFF
--- a/src/processes/generic_process/perturbative/cross_section.jl
+++ b/src/processes/generic_process/perturbative/cross_section.jl
@@ -34,7 +34,7 @@ function QEDbase._matrix_element(psp::PhaseSpacePoint{PROC, PerturbativeQED}) wh
     return sqrt(mat_el_func(psp))
 end
 
-function QEDbase._matrix_element_square(psp::PhaseSpacePoint{PROC, PerturbativeQED}) where {PROC <: ScatteringProcess}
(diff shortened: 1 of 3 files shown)


AntonReinhard:
@SimeonEhrig did something about the integration tests change when using a unit_test CI dependency? I think the docs and QuantumElectrodynamics.jl integration test shouldn't fail.

szabo137:
> @SimeonEhrig did something about the integration tests change when using a unit_test CI dependency? I think the docs and QuantumElectrodynamics.jl integration test shouldn't fail.

Indeed, the environmental variables changed, which is currently only reflected in the dev docs (we should definitely release QuantumElectrodynamics.jl soon). See here: https://qedjl-project.github.io/QuantumElectrodynamics.jl/dev/ci/#Environment-Variables

AntonReinhard:
> Indeed, the environmental variables changed, which is currently only reflected in the dev docs (we should definitely release QuantumElectrodynamics.jl soon). See here: https://qedjl-project.github.io/QuantumElectrodynamics.jl/dev/ci/#Environment-Variables

That's what I used, but it was only working for the unit tests themselves, it seems, not in the docs building job or the integration test...

szabo137 (approved):
Thanks for solving 🙏
LGTM!

